PR #10, opened 2018-09-25
Add country code to production Fastly property.

DFurnes:
I'd accidentally dropped the `X-Fastly-Country-Code` header from the production config, which resulted in users missing their `country` attribute and therefore not getting welcome emails (flagged by Luke & Anthony [in Slack](https://dosomething.slack.com/archives/C03T8SDJW/p1537798336000100)). This affected new user registrations from 9/18 until now.

This pull request adds the `X-Fastly-Country-Code` header to requests in our production config:

screen shot 2018-09-25 at 4 35 26 pm

### DIFF
--- a/dosomething/fastly/main.tf
+++ b/dosomething/fastly/main.tf
@@ -91,6 +91,22 @@ resource "fastly_service_v1" "dosomething" {
     ]
   }
 
+  header {
+    name        = "Country Code"
+    type        = "request"
+    action      = "set"
+    source      = "geoip.country_code"
+    destination = "http.X-Fastly-Country-Code"
+  }
+
+  header {
+    name        = "Country Code (Debug)"
+    type        = "response"
+    action      = "set"
+    source      = "geoip.country_code"
+    destination = "http.X-Fastly-Country-Code"
+  }
+
   request_setting {
     name      = "Force SSL"
     force_ssl = true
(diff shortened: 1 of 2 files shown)


DFurnes:
For later: it'd be interesting to look into any solutions for catching mistakes like this automatically… maybe diffing the two Fastly properties and seeing if one contains rules not found in the other?

sheyd:
I like the diff idea. One thing we might be able to do if Terraform support this is putting the config in place on Fastly as a new staged revision, without promoting it, then doing the diff and promoting separately after a PR?

DFurnes:
> The Country Code header is set twice, once normally, and once as (Debug). Why does it need to be set twice?

Yep, the `"Country Code"` header is a request header, so it's set on the incoming request before it hits our backend (and is what Northstar uses to set `country_code`). The `"Country Code (Debug)"` header is a response header, and so is set on the response to the user (for easier debugging within web inspector or cURL).

sheyd:
I swear I can read. I looked at `response` and `request`, and they looked the same to me when I glanced 10 minutes ago. I need some time away from my screen. :sunglasses:

DFurnes:
> One thing we might be able to do if Terraform support this is putting the config in place on Fastly as a new staged revision, without promoting it, then doing the diff and promoting separately after a PR?

I bet we could probably do this with Terraform's plan/state without having to apply any changes to the actual properties in Fastly. I'll make an issue and we can continue to mull over it!

DFurnes:
Applied this change & confirmed that we're now getting `country` set on new registrations!

PR #18, opened 2018-10-19
Allow Facebook & Twitter to crawl QA/dev sites.

### LINKED ISSUE #16: Need to allow some crawlers to access www-dev, www-qa, www-preview
robots txt blocking crawler

The `robots.txt` on our different dev/test environments is blocking testing out different crawlers from reading our pages to see how cards are previewing on social platforms like Twitter and Facebook.

DFurnes:
This should fix #16 by allowing the `Facebot`, `facebookexternalhit` & `Twitterbot` user-agents to access our QA environments.

Interestingly, Facebook _was_ able to crawl the dev site when I tested without this change, but their documentation does say that they [respect `robots.txt`](https://developers.facebook.com/docs/sharing/webmasters/crawler) so it seems safer to whitelist.

### DIFF
--- a/shared/deny-robots.txt
+++ b/shared/deny-robots.txt
@@ -1,3 +1,14 @@
+# Prevent crawlers, like Google, from
+# indexing our QA & development sites.
 User-agent:*
-
 Disallow: /
+
+# Allow Facebook & Twitter crawlers for testing.
+User-agent: Facebot
+Disallow:
+
+User-agent: facebookexternalhit
+Disallow:
+
+User-agent: Twitterbot
+Disallow:


DFurnes:
Applied this change and tested with Twitter's Card Validator. It works again!

screen shot 2018-10-19 at 11 11 40 am

DFurnes:
Confirmed I'm still able to see this URL when refreshing Facebook's Open Graph Debugger with the new `robots.txt` change & seeing previews in iOS's Messages app:

screen shot 2018-10-19 at 11 19 22 am

PR #27, opened 2018-10-30
Route 'Submit Your Idea' page to Ashes.

DFurnes:
This pull request updates our routing rules for `/us/campaigns` to route `submit-your-idea`, the Member Created Campaign form to Ashes. References [this Slack conversation](https://dosomething.slack.com/archives/C03T8SDJJ/p1540858014007200).


DFurnes:
After some further investigation, we realized this is a redirect created in the Drupal node edit form and wasn't working correctly even if routed to Ashes. I've created a 301 redirect from `/us/campaigns/submit-your-idea` in Aurora & confirmed that it's routing happily now.

Issue #31, opened 2018-08-10
Improve Northstar performance & concurrency.


DFurnes:
This is a follow-up ticket from DoSomething/devops#383.

#### Background
We successfully moved Northstar from AWS (an m4.xlarge EC2 instance) to Heroku (1-5 autoscaled Performance-M dynos). This simplified our operations & also allows us to scale up-and-down based on demand more easily. I've monitored performance and made some adjustments since then:

From [July 17th 2018's broadcast](https://github.com/DoSomething/devops/issues/383#issuecomment-406007425):

> We ran a couple of broadcasts yesterday - a smaller 11:00am broadcast, and a 2:00-8:30pm full-list broadcast (which is continuing today). Both were [set to run at 75rps](https://dosomething.slack.com/archives/C02BBRN5A/p1531504265000143), and seem to have happily [ran at between 70-85rps(!!)](https://dosomething.slack.com/archives/C02BBRN5A/p1531921749000455). Here's what [New Relic](https://rpm.newrelic.com/accounts/108038/applications/92216128) had to say:
> 
> ![screen shot 2018-07-18 at 11 12 06 am](https://user-images.githubusercontent.com/583202/42890869-ec39aae4-8a7b-11e8-9d53-d6f91bcb8188.png)
> 
> And here's that same response time graph overlaid over auto-scaling events:
> 
> ![screen shot 2018-07-18 at 1 09 53 pm](https://user-images.githubusercontent.com/583202/42896738-e17895b0-8a8b-11e8-94a0-ac52f3f31526.png)
> 
> It's interesting to see how much scaling fluctuates between 2-4 dynos when under load, which perhaps suggests we should drop our desired p95 response time a little more (currently at 750ms, curious to try 500ms again).

And [the following day](https://github.com/DoSomething/devops/issues/383#issuecomment-406034442) when we finished that broadcast:

> I think one problem we're running into with auto-scaling is that it's entirely based on response time, so if we get good performance at 3 dynos, we scale down to 2 until things bog down, and then back up... for example, here's the hour from 12:50-1:50pm with a target 500ms p95 response time:
> 
> <img width="485" alt="screen shot 2018-07-18 at 4 59 25 pm" src="https://user-images.githubusercontent.com/583202/42907494-f5837ea6-8aab-11e8-86d0-c63b05ab993d.png">

#### Current Behavior
[...]

PR #32, opened 2018-11-01
WIP: Enable shielding for backend services.

DFurnes:
We're trying to reduce load on Northstar by caching responses in Fastly, but ran into problems since responses are cached per-node by default (and so if we hit a different Fastly POP on the next request, it'll still be a `MISS`).

This pull request enables [shielding](
https://docs.fastly.com/guides/performance-tuning/shielding) for our backend properties so all requests are funneled through a single POP for maximized cache hits (and few downsides, since all our backend services are already located in the same US East region & so don't need global POPs).

References #31.

### DIFF
--- a/dosomething-dev/fastly-backend/main.tf
+++ b/dosomething-dev/fastly-backend/main.tf
@@ -66,6 +66,7 @@ resource "fastly_service_v1" "backends-dev" {
     address           = "${var.graphql_backend}"
     name              = "${var.graphql_name}"
     request_condition = "backend-graphql-dev"
+    shield            = "iad-va-us"
     auto_loadbalance  = false
     port              = 443
   }
@@ -74,6 +75,7 @@ resource "fastly_service_v1" "backends-dev" {
     address           = "${var.northstar_backend}"
     name              = "${var.northstar_name}"
     request_condition = "backend-northstar-dev"
+    shield            = "iad-va-us"
     auto_loadbalance  = false
     port              = 443
   }
@@ -82,6 +84,7 @@ resource "fastly_service_v1" "backends-dev" {
     address           = "${var.phoenix_backend}"
     name              = "${var.phoenix_name}"
     request_condition = "backend-phoenix-dev"
+    shield            = "iad-va-us"
     auto_loadbalance  = false
     port              = 443
   }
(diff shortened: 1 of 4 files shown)


sheyd:
@DFurnes Yay for enabling shielding! This helped a lot when we enabled it at :older_woman: :woman_cook: too.

mshmsh5000:
First check to make sure geo headers make it through. We originally enabled Fastly with shielding, but headers got obliterated (the location of every incoming request was rewritten as the location of the shielding PoP).

If they don't, we can likely figure out some header manipulation to pass them through the shield. (Or, maybe Fastly has addressed this in the last few years.)

DFurnes:
Mmm, good point! I added country code to the Papertrail logs in a33c087, and confirmed that it seems to be recorded correctly with shielding enabled via [a WebPageTest run from France](https://www.webpagetest.org/result/181101_W0_042719df3b69396b8b5ab5b1f1e4a91f/):

```
Nov 01 12:05:51 fastly-qa northstar-qa:  [01/Nov/2018:16:05:50 +0000] 'GET /register HTTP/1.1' status=302 cache=HIT country=FR bytes="-" microseconds=1192
```

mshmsh5000:
Great! Maybe Fastly has just fixed this since we first tried it (which, IIRC, was for Celebs Gone Good quite a while ago).

DFurnes:
And finally, confirming that cache hits are indeed maintained with shielding enabled, even when connecting via different POPs. I think this seems safe to bump up to production once I get a review!

DFurnes:
> Nit-picky, and not blocking question: is there a scenario when there wouldn't be a beresp.http.X-Origin-Name header (referencing https://github.com/DoSomething/infrastructure/pull/32/files#diff-337ac16c7d06457ebee7e95b715e6cca)?

Yes! That header isn't set on the initial hit on the origin POP. I added that conditional in 3441799 because otherwise the header was being re-written to the name of the shield POP, which isn't what we want to see there (and can always see on Fastly's automatic `X-Served-By` header).

sheyd:
Thanks for the explanation. :Fino:

DFurnes:
Alright, I applied this to production and everything seems good! Merging. 🛠

Issue #42, opened 2018-11-02
Create homepage takeover for election day.


DFurnes:
We're preparing a takeover for www.dosomething.org for election day, with resources for our members on how & where to vote. Luke has prepared the markup for this on the [voter-reg](https://github.com/DoSomething/static-rapid-response/tree/election-day) branch of our [static-rapid-response](https://github.com/DoSomething/static-rapid-response) repo. We'll use either a Fastly synthetic response or S3 bucket to host it.

PR #43, opened 2018-11-02
Add election day takeover as synthetic response.

DFurnes:
References #42.


DFurnes:
This is ready at https://qa.dosomething.org/us for testing, and looking good to me! I left the Mexican and Brazilian homepage paths out of the condition since this is a US-specific takeover.

ngjo:
Just went through this and confirming that the only issues I see are known (missing links & the weird highlighting over the buttons) and Luke has done QA w/ the Creative team. I'm going to follow up w/ Tej on Monday for the missing links and to see if any of them should be UTM'd.

How can/should I QA tracking on this page?

dosomething/fastly-frontend/main.tf
```diff
@@ -29,14 +42,12 @@ resource "fastly_service_v1" "frontend" {
     use_ssl           = true
     port              = 443
   }
-
```
DFurnes: This is so funny! Terraform's formatter seems to get confused by the commented out block above and removes spacing between all subsequent parts of this resource. Doesn't seem to cause any issues besides an ugly diff.

DFurnes:
> How can/should I QA tracking on this page?

I don't think we've set up a Google Analytics property for QA, unfortunately (other non-takeover pages on Ashes also reference the production Analytics ID, and so I think any metrics recorded there are discarded by Google). We can definitely test this first thing when we launch on production.

ngjo:
Thanks @DFurnes  - will huddle up to test the analytics once this is on prod. Would it be useful to make these Bertly links to track clicks in case analytics doesn't work as planned tomorrow?

DFurnes:
@ngjo Yeah, that's a good idea! 👍

DFurnes:
Updated with links & style fixes from Luke in 1b6fe7a, and shortened links w/ Bertly in 33142ec.

DFurnes:
Applied Luke's latest changes in f7291b5 to QA!

ngjo:
Looks good to me @DFurnes! I clicked on the bertly links, but it takes a little while for it to show up in Looker - but if there's a faster way for you to validate the clinks are tracking let me know and I can pass my NS ID. I think there's some final text changes that are coming up in the voter reg task force meeting that Luke might be getting after we break.

DFurnes:
I've verified that clicks are coming through in Bertly (by appending `/clicks` to each shortened URL). We won't be able to see individual Northstar IDs clicking through since they're just three shortened links for the static takeover (unlike how we make specific links for each user in a broadcasts).

DFurnes:
Alright, I think I've got a time-based takeover configured with 9523149 & cb5ea63. It's visible (toggled on) right now at https://qa.dosomething.org/us, and should toggle off at 5:03pm... we'll see!

DFurnes:
Whoops, was waiting till 6:03pm because I did a bad time zone conversion. I updated the end time in d9ee8f0, and confirmed it showed the takeover at 5:07pm EST and the standard homepage at 5:08pm!

DFurnes:
And made one last update to do the time zone conversion in code, rather than in my error-prone head in 
201f311. Applied, and confirmed that the takeover was showing at 5:16pm and is back to the normal site at 5:17pm. 🕰

DFurnes:
This should be ready to apply to production (pending a review) and will automatically apply for the duration of November 6th (12:00am-11:59pm EST). I'll be on call to double-check that everything swaps over happily tonight and back tomorrow night. cc: @ngjo

DFurnes:
Alright, applied to production just now. Merging!

PR #44, opened 2018-11-06
Updated button text

lkpttn:
It's simple! afaik there shouldn't be anymore changes 🤞


DFurnes:
I've applied this change & it's live on www.dosomething.org! 🦅 

screen shot 2018-11-06 at 11 38 47 am

DFurnes:
References #42.

PR #45, opened 2018-11-07
Clean up & abstract homepage takeover for future use.

### LINKED ISSUE #42: Create homepage takeover for election day.

DFurnes:
This pull request includes some cleanup for the synthetic homepage takeover we did on Election Day – separating out environment-specific configuration from the shared `RECV` snippet so it's possible to test these out on QA before production without modifying the snippet.

Closes #42.


DFurnes:
I've applied the changes in this PR (commented out on production for now), and confirmed that everything still works on QA in 1202ccf. 🆗

shared/takeover_recv.vcl
```diff
@@ -0,0 +1,11 @@
+# Ensure this value is only set within our VCL:
+unset req.http.X-Timed-Synthetic-Response;
+
+# If we're on the US homepage at the right time, show the takeover:
+if (req.url.path ~ "(?i)^\/((us)\/?)?$" &&
+    # Fastly only deals with times in GMT (so we parse our time in EST and add 5 hours).
+    time.is_after(now, time.add(std.time(var.takeover_start, std.integer2time(-1)), 5h)) &&
```
mshmsh5000: I get the GMT conversion, but what does the `(-1)` do? I see reference to it in Fastly docs too, but I don't get it.

DFurnes: Yeah, it's awkward isn't it? I believe that just creates an "invalid" date, which signals that the fallback parameter for `std.time` is unused (since a compile error is thrown if you assume it's optional and leave that second parameter empty).

DFurnes:
Applied this to production! Merging this in and closing out #42.

PR #50, opened 2018-11-14
Serve Breakup Bash from Phoenix.

DFurnes:
We'd like to start serving [Breakup Bash](https://www.dosomething.org/us/campaigns/breakup-bash) from Phoenix so that we can build a Ghost Inspector monitoring test on a database campaign (that doesn't change routinely). We've [done a round of QA](https://dosomething.slack.com/archives/C03V8B1HA/p1542221366092200) on this campaign with Adam & it's ready to go!


DFurnes:
Applied 14c5dc9 to production, and confirming that it now serves from Phoenix! Merging.

PR #51, opened 2018-11-15
Move Longshot QA environment to Heroku.

DFurnes:
This is ready for review! I left some notes while working on this in the pull request, and a [summary](#issuecomment-439494137) below. I'd like to merge this and do some final testing, and then we can set up the other new instances of this application in a follow-up PR.

References #49.


DFurnes:
Two notes: I had to create the Heroku pipeline by hand since [Terraform does not support setting pipeline owners yet](https://git.io/fpGGA), and created & linked an empty `dosomething-longshot-qa` app because the pipeline was not showing up as importable when it was empty.

DFurnes:
I've created a module that we can use to build out each instance of this application in af0237e, and used it to plan out resources for `longshot-qa`. Since we already have a database for this instance that we want to continue using, I've imported the existing resource into Terraform's state.

DFurnes:
I've created a new `terraform` IAM user (in lieu of granting further permissions to the existing `terraform-readonly` user), with the same read-only access to everything and temporary full access to IAM, RDS, and SQS. (Necessary to create the SQS queue & IAM user for this application, we can look into some better scoping on these permissions in the future.)

I then applied the changes in this PR up through ebdc3b8, which resulted in the following resources being created as defined in code (and the existing RDS instance being opened up to public traffic):

```
module.longshot.heroku_pipeline.longshot
module.longshot.module.longshot-qa.aws_db_instance.database
module.longshot.module.longshot-qa.aws_iam_access_key.queue_user
module.longshot.module.longshot-qa.aws_iam_user.queue_user
module.longshot.module.longshot-qa.aws_sqs_queue.queue
module.longshot.module.longshot-qa.aws_ssm_parameter.database_password
module.longshot.module.longshot-qa.aws_ssm_parameter.database_username
module.longshot.module.longshot-qa.heroku_app.app
module.longshot.module.longshot-qa.heroku_domain.domain
module.longshot.module.longshot-qa.heroku_drain.papertrail
module.longshot.module.longshot-qa.heroku_pipeline_coupling.app
```

DFurnes:
Alright, this is ready for review! 👀 

This pull request adds a configurable "Longshot application" Terraform module, which creates a Heroku app (with domain, Redis addon, logging, and environment variables), RDS instance, S3 bucket, SQS queue, and an app-specific IAM user with limited permissions.

I've then used this template to create a new "Whitelabel" instance, now called [Longshot QA](http://dosomething-longshot-qa.herokuapp.com), in Heroku which connects to the same database as the current production instance.

We should be able to update the HAProxy instance used for this application to point to the new Heroku instance instead, with a new custom domain hosted at [longshot-qa.dosomething.org](https://longshot-qa.dosomething.org).

Once we've confirmed that everything's working as expected, we can set up new parallel environments for Footlocker Internal, Footlocker production, and H&R Block (by just passing different options to the same module) and test & swap each one.

longshot/application/main.tf
```diff
@@ -0,0 +1,217 @@
...
+    MAIL_DRIVER    = "mandrill"
+
+    # Email:
+    EMAIL_NAME      = "${var.email_name}"
+    EMAIL_ADDRESS   = "${var.email_address}"
+    MAIL_HOST       = "smtp.mandrillapp.com"
+    MANDRILL_APIKEY = "${data.aws_ssm_parameter.mandrill_api_key.value}"
+
+    # Database:
+    DB_HOST     = "${aws_db_instance.database.address}"
```
sheyd: @DFurnes Checking that we don't have any special parameter groups for the RDS instance we need to specify here?

longshot/main.tf
```diff
@@ -0,0 +1,24 @@
...
+  host           = "longshot-qa.dosomething.org"
+  pipeline       = "${heroku_pipeline.longshot.id}"
+  pipeline_stage = "staging"
+
+  email_name    = "Longshot QA"
+  email_address = "devops@dosomething.org"
+
+  database_name  = "whitelabel_scholarship"
+  database_type  = "db.t2.micro"
+  database_scale = 5
```
sheyd: Is `database_scale` a Terraformism, or can we control the name of this var?

sheyd: For clarity, the name seems unclear at a glance, and something like `database_storage_size` or something like that would be clearer, I think.

DFurnes: That's a fair point! I'd just used `scale` for consistency with other variables here.

longshot/application/main.tf
DFurnes: Nope, this instance is just using the default parameter & option groups for MariaDB 10.0.

DFurnes: Good to double-check! 👍

longshot/main.tf
DFurnes: Renamed `database_scale` to `database_size` and added a description to clarify!

longshot/main.tf
```diff
@@ -0,0 +1,24 @@
...
+
+resource "heroku_pipeline" "longshot" {
+  name = "longshot"
+}
+
+module "longshot-qa" {
+  source = "application"
+
+  name           = "longshot-qa"
+  host           = "longshot-qa.dosomething.org"
```
katiecrane: this is music to my ears!

katiecrane:
We have tested this with Heroku and nothing jumps out at me as super unclear or missing. Not super familiar with Terraform but looks good to me!

longshot/main.tf
sheyd: 🎸 🔈 🎹

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
+variable "pipeline" {}
```
yzabiyaka: Hey Dave, is this file a copy of some other .tf in DS codebase with minor modifications?

longshot/main.tf
```diff
@@ -0,0 +1,24 @@
...
+  pipeline       = "${heroku_pipeline.longshot.id}"
+  pipeline_stage = "staging"
+
+  email_name    = "Longshot QA"
+  email_address = "devops@dosomething.org"
```
yzabiyaka: who checks this email?

shared/sqs-policy.json.tpl
```diff
@@ -0,0 +1,16 @@
+{
+  "Version": "2012-10-17",
```
yzabiyaka: was "version" this auto-generated or manual?

longshot/main.tf
sheyd: This is an email group that currently consists of Matt, Morgan, Gleb, and myself. Dave, did we add you to this? If not, we should.

DFurnes: I believe the `devops@dosomething.org` address forwards to Matt, Sena, and Morgan. It actually may be worth removing this environment variable in Longshot, since I believe this "from" address is actually set by a value stored in the application's database in most cases.

longshot/application/main.tf
DFurnes: Sort of. I'd previously configured each application instance independently (for example, [northstar](https://github.com/DoSomething/infrastructure/blob/22bbb8a7c0b901dde3269d509589b91ea48525a1/dosomething/northstar/main.tf) and [northstar-dev](https://github.com/DoSomething/infrastructure/blob/master/dosomething-dev/northstar/main.tf)). This is a first stab at standardizing this configuration into reusable modules that we can share amongst environments or similar apps.

longshot/main.tf
DFurnes: Oh, @sheyd beat me to it! Nope, I don't think I'm on this group but may just not have noticed.

katiecrane: @DFurnes we still need an email set in the environment because that is what the password reset via Laravel pulls from

sheyd: 💡

DFurnes: @katiecrane Ahhh, that explains it! 🙇

longshot/application/main.tf
yzabiyaka: General suggestion, not necessary to address in this PR: it is good to branch/clone whatever the existing file so that we get a diff in git. I think this reduces cognitive burden on the reader and in the absence of templates .tf files allows to see which sections are shared among apps. I would imagine one possible option: northstar being copy of northstar-dev.

yzabiyaka: And another nit, once you do create a common template, it might make sense to add comments around parts that are not self-explanatory.

longshot/main.tf
yzabiyaka: Thank you! and hey Kathie! Might be worth adding a comment with Katie's explanation.

shared/sqs-policy.json.tpl
yzabiyaka: ty!

longshot/application/main.tf
yzabiyaka: maybe add a header comment saying this is a template and you expect instantiations of this template to initialize variables you listed in the beginning and didn't assign any value.  It might make sense to group all the unassigned values together, to help the reader out. Also might be good to separate required and optional (with comments)

longshot/main.tf
```diff
@@ -0,0 +1,24 @@
...
+
+  database_name = "whitelabel_scholarship"
+  database_type = "db.t2.micro"
+  database_size = 5
```
yzabiyaka: nit: add a comment with units (GB).

main.tf
```diff
@@ -120,6 +120,21 @@ module "dosomething-dev" {
   papertrail_destination_fastly_dev = "${var.papertrail_destination_fastly_dev}"
```
yzabiyaka: is "_dev" indicates _dev environment? if so, it shouldn't be hardcoded in template and be taken as a var.

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
...
+variable "web_scale" {
+  default = "1"
+}
+
+variable "queue_size" {
+  default = "Standard-1X"
+}
+
+variable "queue_scale" {
+  default = "1"
```
yzabiyaka: are all these defaults assume prod instance? maybe specify what defaults are good for. It might also be good to have separate templates with defaults for different environments: qa, dev, staging, prod.

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
...
+  description = "The amount of storage to allocate to the database, in GB."
+}
+
+variable "email_name" {}
+variable "email_address" {}
+
+variable "papertrail_destination" {}
+
+data "aws_ssm_parameter" "database_username" {
+  name = "/${var.name}/rds/username"
```
yzabiyaka: I don't know enough about this. But will the username always be in rds? if you can see it changing, consider declaring it as a const in the beginning of the file.

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
...
+  name = "/${var.name}/rds/password"
+}
+
+data "aws_ssm_parameter" "mandrill_api_key" {
+  name = "/mandrill/api-key"
+}
+
+resource "heroku_app" "app" {
+  name   = "dosomething-${var.name}"
+  region = "us"
```
yzabiyaka: region should be a parameter for template

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
...
+
+  organization {
+    name = "dosomething"
+  }
+
+  config_vars {
+    # App settings:
+    APP_ENV   = "production"
```
yzabiyaka: env should be a parameter

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
...
+    APP_DEBUG = "false"
```
yzabiyaka: debug should be conditional on env, if prod then false if dev then true.

longshot/application/main.tf
```diff
@@ -0,0 +1,218 @@
...
+  skip_final_snapshot = true
+
+  tags = {
+    Application = "${var.name}"
+  }
+}
+
+resource "aws_sqs_queue" "queue" {
+  name                      = "${var.name}"
+  message_retention_seconds = "${60 * 60 * 24 * 14}"
```
yzabiyaka: if there are constants in tf, might make sense to define the constant in the beginning of the file.

longshot/application/main.tf
DFurnes: We'll likely use the same values here for all instances of this application.

longshot/application/main.tf
DFurnes: Yep! This "data resource" loads the configuration variable from [AWS's Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-paramstore.html), so this can basically be considered a constant.

longshot/application/main.tf
DFurnes: Good idea! I've updated documentation for these in #55.

longshot/main.tf
DFurnes: Renamed this variable to `database_size_gb` for clarity in #55.

main.tf
DFurnes: Good point, renamed to `papertrail_destination_fastly` in both places in #55.

longshot/application/main.tf
DFurnes: I don't think we'll be creating applications in non-US regions anytime soon (even when we re-open global traffic, we'll likely let Fastly handle local POPs for us), so this seems safe to leave as a hard-coded string to me.

longshot/application/main.tf
DFurnes: We want to enforce this setting for `APP_DEBUG` on all instances that are publicly accessible, since Laravel uses this to determine whether to display full stack traces & environment dump on errors. If we set `APP_DEBUG` to "true", we'd give away secrets!

longshot/application/main.tf
DFurnes: Yep, Terraform has [local values](https://www.terraform.io/docs/configuration/locals.html) which act as constants. Since this value won't be used elsewhere, I'm going to just add a comment to this line to better explain it's usage in #55.

longshot/application/main.tf
DFurnes: We've tended to just set this to `production` across the board, but definitely could be useful to use actual environment names here. I've updated this in #55.

DFurnes:
> It all looks good in general, except that the following statement seems alarming " Longshot QA, in Heroku which connects to the same database as the current production instance."

Oh, that was poorly phrased! The newly provisioned database for Longshot QA is the same one that the existing QA instance in EC2 is connected to (so only "production" in the sense that it's the version of this QA application that's in active use, whoops). 🤦‍♂️

PR #52, opened 2018-11-16
Trust proxies on Heroku.

DFurnes:
This pull request sets the `TRUSTED_PROXY_IP_ADDRESSES` environment variable in our Longshot application template, for the middleware added in DoSomething/longshot#918.

### DIFF
--- a/longshot/application/main.tf
+++ b/longshot/application/main.tf
@@ -61,10 +61,11 @@ resource "heroku_app" "app" {
 
   config_vars {
     # App settings:
-    APP_ENV   = "production"
-    APP_DEBUG = "false"
-    APP_LOG   = "errorlog"
-    APP_URL   = "https://${var.host}"
+    APP_ENV                    = "production"
+    APP_DEBUG                  = "false"
+    APP_LOG                    = "errorlog"
+    APP_URL                    = "https://${var.host}"
+    TRUSTED_PROXY_IP_ADDRESSES = "**"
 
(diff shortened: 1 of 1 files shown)


DFurnes:
Applied 5df08e5 to Longshot QA & confirming that HTTPS works properly now.

longshot/application/main.tf
```diff
@@ -61,10 +61,11 @@ resource "heroku_app" "app" {
...
     # App settings:
-    APP_ENV   = "production"
-    APP_DEBUG = "false"
-    APP_LOG   = "errorlog"
-    APP_URL   = "https://${var.host}"
+    APP_ENV                    = "production"
+    APP_DEBUG                  = "false"
+    APP_LOG                    = "errorlog"
+    APP_URL                    = "https://${var.host}"
+    TRUSTED_PROXY_IP_ADDRESSES = "**"
```
katiecrane: does this mean we trust everything?

DFurnes: Yup, since we don't know the IP addresses for Heroku's load balancer. If we were hosting this behind a service with a known IP address (like we used to do in AWS), we could have set that here instead.

sheyd: And if so, would having the HA Proxy IP address be listed here be a better solution? Or is this a requirement for HTTPS to work properly at all?

DFurnes:
Thanks @sheyd! 🏹

PR #53, opened 2018-11-16
WIP: Add production Longshot instances to Terraform/Heroku.

DFurnes:
This pull request will add new production instances for Footlocker Internal, Footlocker, and H&R Block (using the same databases as their existing EC2 counterparts, but new Heroku/S3/SQS resources).

### DIFF
--- a/longshot/main.tf
+++ b/longshot/main.tf
@@ -23,7 +23,7 @@ module "longshot-qa" {
   papertrail_destination = "${var.papertrail_qa_destination}"
 }
 
-module "longshot-footlocer-internal" {
+module "longshot-footlocker-internal" {
   source = "application"
 
   name           = "longshot-footlocker-internal"
--- a/longshot/main.tf
+++ b/longshot/main.tf
@@ -22,3 +22,20 @@ module "longshot-qa" {
 
   papertrail_destination = "${var.papertrail_qa_destination}"
 }
+
+module "longshot-footlocer-internal" {
+  source = "application"
+
+  name           = "longshot-footlocker-internal"
+  host           = "footlocker-internal.dosomething.org"
+  pipeline       = "${heroku_pipeline.longshot.id}"
+  pipeline_stage = "production"
+
+  email_name    = "Foot Locker Scholar Athletes"
+  email_address = "footlocker@tmiagency.org"
+
+  database_name = "footlocker_internal"
+  database_type = "db.t2.medium"
+
+  papertrail_destination = "${var.papertrail_prod_destination}"
+}


DFurnes:
Partially applied these resources, but was foiled by this error:

```
Error: Error applying plan:

1 error(s) occurred:

* module.longshot.module.longshot-footlocker-internal.aws_db_instance.database: 1 error(s) occurred:

* aws_db_instance.database: Error creating DB Instance: InvalidParameterCombination: Non-VPC DB Instances are not available for engine: mariadb
	status code: 400, request id: 9b3ff670-4d51-4553-a083-88848d568c87
```

Unfortunately, we don't have a default VPC (which, otherwise the RDS instance would be created in) and [may not be able to create one](https://forums.aws.amazon.com/thread.jspa?messageID=814102) without removing old "incompatible" resources on our account.

Will pick this back up on Monday! 📆

DFurnes:
Alright, was successfully able to create the RDS instance with the change in 09f593f.

One little blip is that Terraform is [unable to change an existing instance's identifier](https://git.io/fpWkO) without re-creating it from scratch, so we can't enforce naming of these instances _and_ import existing named resources with the same template. We _can_ still rename via the AWS UI.

Next up, it turns out `dosomething-longshot-footlocker-internal` is a pretty long name!

```
Error: Error applying plan:

1 error(s) occurred:

* module.longshot.module.longshot-footlocker-internal.heroku_app.app: 1 error(s) occurred:

* heroku_app.app: Post https://api.heroku.com/teams/apps: Name is too long (maximum is 30 characters)
```

I'm going to remove the `dosomething-` prefix from this application, and we can maybe consider whether it's worth changing that to something shorter like `ds-` across the board...

DFurnes:
Was able to create the "footlocker-internal" Heroku app with the change in 2c92403, and fixed up an issue with RDS network access in 2c9873d (the default security group doesn't allow external traffic).

I'm wondering if the best approach here might be to create all new RDS instances with `id` & `name` managed by Terraform, instead of "importing" the existing resources. This would help ensure we don't miss specifying anything in our Terraform config... for example, just noticed I need to enable automated backups on this new instance as well.

sheyd:
`I'm wondering if the best approach here might be to create all new RDS instances with id & name managed by Terraform, instead of "importing" the existing resources. This would help ensure we don't miss specifying anything in our Terraform config... for example, just noticed I need to enable automated backups on this new instance as well.`

I'm a fan of this. We can do an off-hours one time dump/restore with the app offline to ensure data consistency. Maybe also worth moving everything (if it isn't already) to MariaDB or MySQL? whatever standard we want to use if they aren't consistent already.

sheyd:
Generally if we can rebuild with consistent naming and import data with little or one-time downtime, I think it's worth it to do that instead of trying to import existing resources that aren't named to our standards.

DFurnes:
Alright, I've copied over the old database & files from the `footlocker-internal` EC2 instance to this new Terraform-provisioned application, and it's [looking good](https://longshot-footlocker-internal.herokuapp.com)! 🆗

DFurnes:
I've increased the Footlocker Internal RDS instance size (to 100GB) in b094151 to match our other production instances, and provisioned a new production environment for Footlocker in baf1d61.

Here's the current process for creating a new instance:
1. Create `/{name}/rds/username` and `/{name}/rds/password` in [AWS SSM](https://console.aws.amazon.com/systems-manager/parameters/?).
2. Run `terraform apply` to create the new module (see baf1d61). This takes ~10 minutes.
3. You'll run into an expected error creating formations, since the Heroku app doesn't have any code deployed yet. Deploy master to the instance from [the Heroku app's control panel](https://dashboard.heroku.com/apps/longshot-footlocker/deploy/github) (not exposed via Heroku CLI or Terraform resource).
4. Re-run `terraform apply` to configure the app's `web` and `queue` formations.

DFurnes:
Finally, added a new production H&R Block application in 08a53dd. As with the others, this is entirely separate from our old AWS instances & will have no impact on production traffic until we swap over the database & HAProxy backend.

DFurnes:
@sheyd @blisteringherb @mshmsh5000 I'd love a second set of eyes on the new databases that have been provisioned for these apps. Are there any settings missing or configured incorrectly? (Note that for `longshot-hrblock` and `longshot-footlocker`, I have _not_ copied over the database contents yet.)

| application | old database | new database
| ----------- | ------------- | --------------
| longshot-qa (whitelabel) | – | [longshot-qa](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstance:id=longshot-qa)
| longshot-footlocker-internal | – | [longshot-footlocker-internal](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstance:id=longshot-footlocker-internal)
| longshot-footlocker | [footlocker-prod](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstance:id=footlocker-prod) | [longshot-footlocker](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstance:id=longshot-footlocker)
| longshot-hrblock | [hrblock](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstance:id=hrblock) | [longshot-hrblock](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstance:id=longshot-hrblock)

DFurnes:
I've also staged the new database credentials on the `footlocker` and `hrblock` EC2 instances (commented out in each app's `.env`) so we can swap these out before cutting over the actual application servers.

sheyd:
@DFurnes Can we upgrade the engine version to 10.3 here: https://github.com/DoSomething/infrastructure/pull/53/files#diff-e362674ccb7459864648ad25165f049eR157. 10.0 is an oooooooold version of MariaDB, and the upgrades should be fairly benign. Obviously, we can test ton longshot-qa before the others as a test, but I'd love to roll in these changes upfront to avoid downtimes later.

DFurnes:
@sheyd Yes! I've made that change and applied (will take effect in next maintenance window on Friday). It seems like we have to upgrade to 10.2.x first, and then to 10.3.x from there.

screen shot 2018-11-21 at 10 04 34 am

Issue #48, opened 2018-11-09
Capture Quasar Config in Terraform


sheyd:
- [x] Quasar QA DB, Security Groups, and Options Group
- [x] Quasar Prod DB, Security Groups, and Options Group

sheyd:
Adding to this parameter group settings, since we've migrated to RDS via https://www.pivotaltracker.com/story/show/161654902, and one thing we lost was the Heroku auto-tuning they do on their end. 

We have to do some performance tweaks to RDS instance via https://www.pivotaltracker.com/story/show/162208643.

DFurnes:
@sheyd Is this good to close out? 🎉

sheyd:
🎆 🍾

PR #57, opened 2018-11-27
Quasar Setup

sheyd:
Adding basic Quasar setup to Terraform!!!

The basic scaffolding we have here are:
1. Adding the Terraform module for Quasar.
2. Defining VPC and RDS subnets.

These have been imported from AWS Quasar, so there shouldn't be any change diffs that need to made, or any changes made to production or QA.

Dave and I double checked this with a `make plan` before the latest commit, to ensure no changes!


DFurnes:
References #48.

Issue #28, opened 2018-10-30
Standardize more common patterns with modules.


DFurnes:
We tend to duplicate a lot of the same boilerplate for each service (e.g. [Rogue](https://github.com/DoSomething/infrastructure/blob/master/dosomething/rogue/main.tf) and [Northstar](https://github.com/DoSomething/infrastructure/blob/master/dosomething/northstar/main.tf) look pretty much identical), and same for the same app between environments (see [Northstar Dev](https://github.com/DoSomething/infrastructure/blob/master/dosomething-dev/northstar/main.tf), [Northstar QA](https://github.com/DoSomething/infrastructure/blob/master/dosomething-qa/northstar/main.tf), and [Northstar Production](https://github.com/DoSomething/infrastructure/blob/master/dosomething/northstar/main.tf)).

We should consider standardizing some of these common patterns with reusable DoSomething-specific modules (like a `heroku_php_app` module that sets up the app, buildpack, standard environment variables, domain, log drain, and just has arguments for things we may customize per-app).

DFurnes:
We could also consider using versioned [private modules](https://www.terraform.io/docs/enterprise/registry/using.html) if we're able to get a discount on Terraform Enterprise, point at a [tagged git revision](https://www.terraform.io/docs/modules/sources.html#generic-git-repository) of a module in this repository, or just have multiple "manual" revisions as we iterate on things (e.g. a `heroku_php_app/v1` and `heroku_php_app/v2` in this repo).

DFurnes:
I've created a self-contained [application module](https://github.com/DoSomething/infrastructure/blob/6522db3576ed5e726fc04f15594f741edb74aad5/longshot/application/main.tf) for Longshot, which we can then use to [provision all four apps consistently](https://github.com/DoSomething/infrastructure/blob/6522db3576ed5e726fc04f15594f741edb74aad5/longshot/main.tf) (and easily spin up new instances if we ever want to)! The next step here is to extract these details into a generic "laravel application" and "RDS database", etc. modules.

DFurnes:
I've created re-usable Heroku app, MariaDB database, SQS queue, S3 bucket, and IAM user modules in #69, and will be rolling those out across the board over the next couple of sprints.

DFurnes:
I've used these base modules to provision new infrastructure for Northstar, Rogue, and Phoenix in #96. Next, I need to copy over any data to the newly provisioned resources & swap apps over to each one.

[...]

PR #59, opened 2018-11-28
Add New Relic to Footlocker & H&R Block apps.

DFurnes:
This pull request adds optional New Relic support to the Longshot application module. If the `with_newrelic` argument is set, a New Relic Heroku add-on & appropriate environment variables will be set for that application.

References #49.

### DIFF
--- a/longshot/application/main.tf
+++ b/longshot/application/main.tf
@@ -152,8 +152,8 @@ resource "heroku_app" "app" {
     S3_BUCKET         = "${aws_s3_bucket.storage.id}"
 
     # New Relic:
+    NEW_RELIC_ENABLED  = "${var.with_newrelic ? "true" : "false"}"
     NEW_RELIC_APP_NAME = "${var.with_newrelic ? var.name : ""}"
-    NEW_RELIC_LOG      = "${var.with_newrelic ? "stdout" : ""}"
 
     # We can't use a ternary on an optional resource, hence this hack! https://git.io/fp2pg
     NEW_RELIC_LICENSE_KEY = "${join("", data.aws_ssm_parameter.newrelic_api_key.*.value)}"
@@ -205,15 +205,6 @@ resource "heroku_pipeline_coupling" "app" {
 resource "heroku_addon" "redis" {
   app  = "${heroku_app.app.name}"
   plan = "heroku-redis:${var.redis_type}"
-}
-
-# Attach New Relic addon if enabled for this app. This
-# installs the New Relic PHP agent, and we use environment
-# variables on the app to connect it to our org account.
-resource "heroku_addon" "newrelic" {
-  count = "${var.with_newrelic ? 1 : 0}"
-  app   = "${heroku_app.app.name}"
-  plan  = "newrelic:wayne"
 }
 
 resource "aws_db_instance" "database" {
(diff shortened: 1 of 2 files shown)


DFurnes:
Going to test this out on H&R Block first in 9eea907:

```
+ module.longshot.module.hrblock.heroku_addon.newrelic
    id:              "<computed>"
    app:             "longshot-hrblock"
    config_vars.#:   "<computed>"
    name:            "<computed>"
    plan:            "newrelic:wayne"
    provider_id:     "<computed>"

~ module.longshot.module.hrblock.heroku_app.app
    config_vars.0.%:                       "24" => "27"
    config_vars.0.NEW_RELIC_APP_NAME:      "" => "longshot-hrblock"
    config_vars.0.NEW_RELIC_LICENSE_KEY:   "" => "******** SECRET ********"
    config_vars.0.NEW_RELIC_LOG:           "" => "stdout"

~ module.longshot.module.longshot-footlocker-internal.heroku_app.app
    config_vars.0.%:                       "24" => "27"

~ module.longshot.module.longshot-footlocker.heroku_app.app
    config_vars.0.%:                       "24" => "27"

~ module.longshot.module.longshot-qa.heroku_app.app
    config_vars.0.%:                       "24" => "27"
```

longshot/application/main.tf
```diff
@@ -81,6 +84,11 @@ variable "database_size_gb" {
   default     = 100
 }
 
+variable "with_newrelic" {
```
sheyd: Thought from something in you mentioned in another PR related to this issue the other day. Is this module maybe better put in `shared` module, if we want to use it across properties? Or are we not sure we want to stick with New Relic, and this is our first test to see how we like it on Heroku?

DFurnes: Sorry, which module are you referring to here? The overall `longshot/application` module?

DFurnes: Chatted this through IRL. For anyone else who's following along – the `with_newrelic` variable here is just an argument for the module (think of it as a parameter you'd pass a function in other languages), so it's inherently tied to this module and isn't something we could reasonably extract.

We could definitely extract pieces of this `longshot/application` module into more reusable chunks for elsewhere (so say, bundling the `aws_ssm_parameter` & `heroku_addon` into a `shared/heroku-newrelic` module that exports the expected config values) and I'm planning on tackling that further in #28 after this app is wrapped up.

DFurnes:
Hmm, frustratingly the New Relic add-on doesn't install the agent until you complete a build (confirmed by running `php -i` on longshot-hrblock). I think this hasn't bitten us before since we have the New Relic agent on QA instances where we create Heroku builds, but does pose problem here if we want to avoid installing (and paying for) New Relic on QA here.

One solution is to add `ext-newrelic` to this application's `composer.json` (and that actually may be closer to what we want since it doesn't auto-provision a weird parallel Heroku New Relic account for each app), or we could always include the add-on but only set our "real" account on prod...

DFurnes:
I'm leaning towards using the New Relic extension (and adding it to our Homestead's via an `after.sh` step) since that doesn't create the weird parallel accounts & clobber environment variables.

sheyd:
This Approach Approved by 
download

sheyd:
YAAAAS!

PR #62, opened 2018-11-29
Setup Quasar Security Groups

sheyd:
Sets up Quasar VPC security groups for RDS and EC2 instances.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -29,12 +29,18 @@ resource "aws_security_group" "bastion" {
   description = "22 from Everywhere"
   vpc_id      = "${aws_vpc.vpc.id}"
 
-  ingress {
-    from_port   = 22
-    to_port     = 22
-    protocol    = "tcp"
-    cidr_blocks = ["0.0.0.0/0"]
+  tags {
+    Name = "Quasar-Bastion"
   }
+}
+
+resource "aws_security_group_rule" "bastion" {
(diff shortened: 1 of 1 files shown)


quasar/main.tf
```diff
@@ -23,3 +23,188 @@ resource "aws_subnet" "subnet-b" {
...
   }
 }
+
+resource "aws_security_group" "bastion" {
+  name        = "Quasar-Bastion"
+  description = "22 from Everywhere"
+  vpc_id      = "${aws_vpc.vpc.id}"
+
+  tags {
+    Name = "Quasar-Bastion"
```
DFurnes: I wonder if there's a good way to take `name` on resources that have it and automatically create a tag... 🤔

quasar/main.tf
```diff
@@ -23,3 +23,188 @@ resource "aws_subnet" "subnet-b" {
...
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
+}
+
+resource "aws_security_group_rule" "bastion-egress" {
+  security_group_id = "${aws_security_group.bastion.id}"
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = "-1"
```
DFurnes: This means all protocols? Should we make an inline comment explaining that?

DFurnes:
References #48.

PR #67, opened 2018-11-30
Quasar RDS Param Group

sheyd:
Started config for RDS parameter groups.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -223,3 +223,33 @@ resource "aws_db_parameter_group" "quasar-qa" {
     value = "500"
   }
 }
+
+data "aws_ssm_parameter" "qa_username" {
+  name = "/quasar-qa/rds/username"
+}
+
+data "aws_ssm_parameter" "qa_password" {
+  name = "/quasar-qa/rds/password"
+}
+
+resource "aws_db_instance" "quasar-qa" {
+  allocated_storage     = 1000
+  storage_type          = "gp2"
+  engine                = "postgres"
+  engine_version        = "10.5"
+  instance_class        = "db.m5.2xlarge"
+  name                  = "quasar"
+  username              = "${data.aws_ssm_parameter.qa_username.value}"
+  password              = "${data.aws_ssm_parameter.qa_password.value}"
+  parameter_group_name  = "${aws_db_parameter_group.quasar-qa.id}"
+  deletion_protection   = "true"
+  storage_encrypted     = "true"
+  copy_tags_to_snapshot = "true"
+  monitoring_interval   = "60"
+  publicly_accessible   = "true"
+  skip_final_snapshot   = "true"
+
+  tags {
+    workload-type = "other"
+  }
+}
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -214,7 +214,7 @@ resource "aws_db_parameter_group" "quasar-qa" {
   family = "postgres10"
 
   parameter {
-    name  = "rds_logical_replication"
+    name  = "rds.logical_replication"
     value = "1"
   }
 
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -209,7 +209,7 @@ resource "aws_security_group_rule" "rds-egress" {
   cidr_blocks       = ["0.0.0.0/0"]
 }
 
-resource "aws_db_parameter_group" "quasar" {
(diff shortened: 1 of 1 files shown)


quasar/main.tf
```diff
@@ -208,3 +208,48 @@ resource "aws_security_group_rule" "rds-egress" {
...
+  name = "/quasar-qa/rds/username"
+}
+
+data "aws_ssm_parameter" "qa_password" {
+  name = "/quasar-qa/rds/password"
+}
+
+resource "aws_db_instance" "quasar-qa" {
+  allocated_storage     = 1000
+  storage_type          = "gp2"
```
DFurnes: What's `gp2`? Would it be worth adding a comment here?

quasar/main.tf
```diff
@@ -208,3 +208,48 @@ resource "aws_security_group_rule" "rds-egress" {
...
+  name                  = "quasar"
+  username              = "${data.aws_ssm_parameter.qa_username.value}"
+  password              = "${data.aws_ssm_parameter.qa_password.value}"
+  parameter_group_name  = "${aws_db_parameter_group.quasar-qa.id}"
+  deletion_protection   = "true"
+  storage_encrypted     = "true"
+  copy_tags_to_snapshot = "true"
+  monitoring_interval   = "60"
+  publicly_accessible   = "true"
+  skip_final_snapshot   = "true"
```
DFurnes: Do we want this to be `false` for if we do ever delete this?

quasar/main.tf
```diff
@@ -208,3 +208,48 @@ resource "aws_security_group_rule" "rds-egress" {
...
+
+  tags {
+    workload-type = "other"
```
DFurnes: What's this tag used for?

quasar/main.tf
sheyd: It's default config, so removing this line.

quasar/main.tf
sheyd: 🤔 yeah, probably, even though it'll change the value.

quasar/main.tf
sheyd: Honestly, no clue, but I don't remember adding it to the RDS instance creation, so in case something referred to it, I left it in just in case.

DFurnes: Looks like it's [automatically added by AWS](https://forums.aws.amazon.com/thread.jspa?threadID=208410). If we're not using it for anything, let's remove.

quasar/main.tf
```diff
@@ -208,3 +208,42 @@ resource "aws_security_group_rule" "rds-egress" {
...
+
+data "aws_ssm_parameter" "qa_username" {
+  name = "/quasar-qa/rds/username"
+}
+
+data "aws_ssm_parameter" "qa_password" {
+  name = "/quasar-qa/rds/password"
+}
+
+resource "aws_db_instance" "quasar-qa" {
```
DFurnes: Do we need to set [`vpc_security_group_ids`](https://www.terraform.io/docs/providers/aws/r/db_instance.html#vpc_security_group_ids) on this instance with the security groups you created above?

sheyd: Security groups are for 🐸. All ports open all the time!

Good catch, and added it into config via latest commit.

quasar/main.tf
```diff
@@ -208,3 +208,43 @@ resource "aws_security_group_rule" "rds-egress" {
...
+  allocated_storage      = 1000
+  engine                 = "postgres"
+  engine_version         = "10.5"
+  instance_class         = "db.m5.2xlarge"
+  name                   = "quasar"
+  username               = "${data.aws_ssm_parameter.qa_username.value}"
+  password               = "${data.aws_ssm_parameter.qa_password.value}"
+  parameter_group_name   = "${aws_db_parameter_group.quasar-qa.id}"
+  vpc_security_group_ids = ["${aws_security_group.rds.id}"]
+  deletion_protection    = "true"
```
DFurnes: One last nit-pick – let's use actual booleans for these even though Terraform doesn't care:

```suggestion
  deletion_protection    = true
```

sheyd: `Suggested change Beta` boolean update complete.

sheyd:
@DFurnes Can you give this a final 👍 so I can merge?

DFurnes:
References #48.

PR #83, opened 2018-12-12
Route 'batch' path to Ashes for Solr re-indexing.

DFurnes:
This pull request routes `us/batch/` to Ashes [so we can re-index Solr](https://dosomething.slack.com/archives/C2BPA7M8F/p1544642232032700?thread_ts=1544039437.060400&cid=C2BPA7M8F).


DFurnes:
Applied and confirmed that this URL is now routing to Ashes. 📡

PR #86, opened 2018-12-13
Removed missing campaign: "Cool Carol Crew".

mirie:
This campaign is supposed to be served from Phoenix, and was missing from the
last batch that was switched over from Ashes.


DFurnes:
Applied 05a0326 and confirmed that [Cool Carol Crew](https://www.dosomething.org/us/campaigns/cool-carol-crew) is now serving from Phoenix!

PR #88, opened 2018-12-13
Route about pages to Phoenix by default (QA only).

mirie:
Updates the lookup table to hold exceptions for about pages that should be served from Ashes for QA. The majority of pages should be served from Phoenix now.

### DIFF
--- a/dosomething-dev/fastly-frontend/ashes_recv.vcl
+++ b/dosomething-dev/fastly-frontend/ashes_recv.vcl
@@ -27,8 +27,8 @@ else if (req.url.path ~ "(?i)\/((us|mx|br)\/)?(fact|sobre|volunteer|voluntario|r
   set req.http.X-Fastly-Backend = "ashes";
 }
 else if (req.url.path ~ "(?i)\/((us|mx|br)\/)?about/([A-Za-z0-9_\-]+)" &&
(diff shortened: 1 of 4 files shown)


mirie:
I only updated the QA files - should I also update Dev too?

mirie:
Okay @DFurnes: I also updated dev too!

dosomething-dev/fastly-frontend/ashes_recv.vcl
```diff
@@ -27,8 +27,8 @@ else if (req.url.path ~ "(?i)\/((us|mx|br)\/)?(fact|sobre|volunteer|voluntario|r
   set req.http.X-Fastly-Backend = "ashes";
 }
 else if (req.url.path ~ "(?i)\/((us|mx|br)\/)?about/([A-Za-z0-9_\-]+)" &&
-    ! table.lookup(phoenix_about, std.tolower(re.group.3))) {
-  # About pages default to Ashes, but we'll opt some paths to Phoenix.
+    table.lookup(phoenix_about, std.tolower(re.group.3))) {
```
DFurnes: I think this needs to be `ashes_about` to match the change in `ashes_init.vcl` 👾

DFurnes:
Applied, and confirming things look good on QA & development! 🎉

PR #89, opened 2018-12-14
Route about pages to Phoenix by default (Prod)

mirie:
Updates the lookup table to hold exceptions for about pages that should be served from Ashes for Prod. The majority of pages should be served from Phoenix now.

Checked the pages (reference list = https://docs.google.com/spreadsheets/d/1Z9OGap90agKz9ppvmj1GzWSJaPLdKkHBr1rS1hGs6QM/edit#gid=0) on QA. Looks good, including the exception! So making the same changes for Prod now.

### DIFF
--- a/dosomething/fastly-frontend/ashes_init.vcl
+++ b/dosomething/fastly-frontend/ashes_init.vcl
@@ -305,9 +305,9 @@ table phoenix_facts {
   # ...
 }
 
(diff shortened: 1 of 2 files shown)


DFurnes:
Applied b6498c7 and confirmed that things look good on production. Merging! 🎉 🚀 🎊

PR #91, opened 2018-12-14
Add PASS cache setting to backend requests with auth token.

### LINKED ISSUE #90: Do not serve cached edge content to authenticated requests.
We added caching to Northstar's user profile endpoint in [an attempt to improve broadcast performance](https://github.com/DoSomething/infrastructure/issues/31). Rafa flagged [the other day](https://goo.gl/36cAun) that he was seeing unexpected caching on authenticated requests too, though, which is not expected behavior.

From some futher investigation this morning, I've confirmed that once a profile is cached in Fastly, we will continue to return that cached "public" profile, even for requests that come in with a privileged authentication token. That's not helpful!

To fix this, we should add a ["pass" cache setting](https://docs.fastly.com/guides/vcl-tutorials/understanding-the-different-pass-action-behaviors) for requests with an `Authorization` header.

DFurnes:
This should address an issue where we'd continue to serve cached content (e.g. public profiles where we'd set a `Cache-Control` header) to subsequent authenticated requests. Fixes #90.

### DIFF
--- a/dosomething-dev/fastly-backend/main.tf
+++ b/dosomething-dev/fastly-backend/main.tf
@@ -96,10 +96,10 @@ resource "fastly_service_v1" "backends-dev" {
     statement = "req.http.Authorization"
   }
 
-  cache_setting {
-    name            = "pass-authenticated"
-    cache_condition = "is-authenticated"
-    action          = "pass"
+  request_setting {
+    name              = "pass-authenticated"
+    request_condition = "is-authenticated"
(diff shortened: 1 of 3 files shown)


DFurnes:
Alright, this is working as expected now on QA – anonymous requests continue to be cached in Fastly & authenticated requests will always hit the backend even if the same URL has a cached unauthenticated response. Ready for review! 👁

PR #96, opened 2018-12-17
Add database, storage, queue, and IAM resources to apps.

DFurnes:
This pull request adds database, storage, queue, and IAM role resources to Northstar, Rogue, and Phoenix so we can manage and track these with Terraform. The game plan will be to provision the new resources (which use new, standardized environment variables & resource names) and we can swap to use them in production as we test each newly provisioned resource.

**Note:** None of this is applied, just pushing up my work before I'm OOO for the rest of the week!

### DIFF
--- a/applications/longshot/main.tf
+++ b/applications/longshot/main.tf
@@ -88,6 +88,7 @@ module "database" {
   source = "../../shared/mariadb_instance"
 
   name              = "${var.name}"
+  database_name     = "longshot"
   instance_class    = "${var.environment == "production" ? "db.t2.medium" : "db.t2.micro"}"
   allocated_storage = "${var.environment == "production" ? 100 : 5}"
 
--- a/applications/longshot/main.tf
+++ b/applications/longshot/main.tf
@@ -88,6 +88,7 @@ module "database" {
   source = "../../shared/mariadb_instance"
 
   name              = "${var.name}"
+  environmnet       = "${var.environment}"
   database_name     = "longshot"
   instance_class    = "${var.environment == "production" ? "db.t2.medium" : "db.t2.micro"}"
   allocated_storage = "${var.environment == "production" ? 100 : 5}"
--- a/applications/longshot/main.tf
+++ b/applications/longshot/main.tf
@@ -88,7 +88,7 @@ module "database" {
   source = "../../shared/mariadb_instance"
 
   name              = "${var.name}"
-  environmnet       = "${var.environment}"
+  environment       = "${var.environment}"
   database_name     = "longshot"
   instance_class    = "${var.environment == "production" ? "db.t2.medium" : "db.t2.micro"}"
   allocated_storage = "${var.environment == "production" ? 100 : 5}"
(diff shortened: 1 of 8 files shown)


DFurnes:
Rebased to snag the latest from `master`. The next step here is to see if I can set up database replication to migrate data over to these newly provisioned MariaDB instances without any manual work.

DFurnes:
I think it's probably worth reviewing and applying this now, since the plan is actually quite large. As of now, this should provision new RDS, SQS, S3, and IAM resources for all these applications __but not swap each app to actually connect to the new resource until we're ready__.

To do so, I've either left environment variables untouched or applied the new values under previously unused names (e.g. so we can start using a standardized `AWS_ACCESS_KEY` across the board instead of Rogue's existing `S3_KEY` & `SQS_ACCESS_KEY_ID`, or Northstar's `S3_KEY` & `SQS_PUBLIC_KEY`).

DFurnes:
Noticed two more little things with this:

🔏 I had previously defaulted these database instances to the TMI VPC, which doesn't make sense outside of Longshot. I've swapped the default to the `DoSomething` VPC in 1095bd6, and we can consider updating our account from EC2-Classic so we can just have this default in the future.

📛 Removes hardcoded `longshot` database name in c43c514. It now defaults to the application name (e.g. `dosomething-rogue-qa`) or allows a `database_name` argument to override in Terraform.

DFurnes:
Hah, always something! Funny since this worked fine with SequelPro on my local. 

```
Error creating DB Instance: InvalidParameterValue: DBName must begin with a letter and contain only alphanumeric characters.
```

I've updated these to be `snake_case` in e650276.

DFurnes:
Two more tiny issues that popped up when applying this – typo'd the queue export name and fixed in 
2028464, and had to hard-code the default security group in 97160bb since [HCL doesn't currently hav a concept of `null`](https://git.io/fhYmP).

DFurnes:
I'm sanity-checking things against their existing counterparts. Noticed that we did not set `multi_az` on the new Rogue production instance, and added in 5f85811. (Related to this [Slack convo](https://dosomething.slack.com/archives/C03T8SDJJ/p1546535132013400).)

DFurnes:
I also noticed that Phoenix production seems to be over-provisioned with a `db.t2.large`, since it only uses the database for storing authentication tokens & a deprecated referral feature. I've scaled down to a `db.t2.medium` in f502c33 and we can consider tuning further in the future.

DFurnes:
I've updated the MariaDB template to increase the automatic backup retention period to 30 days, and to disable automatic backups on QA/development instances since they aren't useful there. I also enabled the slow query log across the board, as that can be helpful debugging performance issues.

I didn't end up enabling [enhanced monitoring](https://goo.gl/xMx7GD) on these new databases due to a [Terraform order-of-operations bug](https://git.io/fhYuc) w/ creating the monitoring role & attaching it. I think we get most (or all) of the metrics we want without this, so always something we can revisit later.

DFurnes:
I'm going to merge this in & continue work on these in a follow-up PR.

PR #99, opened 2019-01-03
Use newly provisioned resources for Northstar, Phoenix & Rogue.

DFurnes:
This pull request continues the work started in #96 by swapping the applications to actually start using each newly provisioned resource (once we've copied over any necessary data).

In a shocking twist, Terraform [doesn't yet support maps in ternaries](https://git.io/fhYr7) (or any other sort of conditional) and so there's no easy way to only apply these changes to dev & QA environments via Terraform. Jeez! Since the resources already exist, I'm going to use the results of the plan to manually test each of the config changes on QA before applying across the board.

### DIFF
--- a/shared/s3_bucket/main.tf
+++ b/shared/s3_bucket/main.tf
@@ -13,6 +13,11 @@ variable "acl" {
   default     = "public-read"
 }
 
+variable "force_public" {
+  description = "Force 'public read' permissions for objects. Not recommended."
+  default     = false
+}
+
 resource "aws_s3_bucket" "bucket" {
   bucket = "${var.name}"
   acl    = "${var.acl}"
@@ -23,7 +28,7 @@ resource "aws_s3_bucket" "bucket" {
 }
 
 data "template_file" "s3_policy" {
-  template = "${file("${path.module}/iam-policy.json.tpl")}"
+  template = "${file("${path.module}/${var.force_public ? "public-" : ""}iam-policy.json.tpl")}"
 
   vars {
     bucket_arn = "${aws_s3_bucket.bucket.arn}"
--- a/shared/s3_bucket/main.tf
+++ b/shared/s3_bucket/main.tf
@@ -28,7 +28,22 @@ resource "aws_s3_bucket" "bucket" {
 }
 
 data "template_file" "s3_policy" {
-  template = "${file("${path.module}/${var.force_public ? "public-" : ""}iam-policy.json.tpl")}"
+  template = "${file("${path.module}/iam-policy.json.tpl")}"
+
+  vars {
+    bucket_arn = "${aws_s3_bucket.bucket.arn}"
+  }
+}
+
+resource "aws_s3_bucket_policy" "bucket_policy" {
+  count = "${var.force_public ? 1 : 0}"
+
+  bucket = "${aws_s3_bucket.bucket.id}"
+  policy = "${data.template_file.public_bucket_policy.rendered}"
+}
+
+data "template_file" "public_bucket_policy" {
+  template = "${file("${path.module}/public-bucket-policy.json.tpl")}"
 
   vars {
     bucket_arn = "${aws_s3_bucket.bucket.arn}"
--- a/applications/northstar/main.tf
+++ b/applications/northstar/main.tf
@@ -88,15 +88,6 @@ locals {
     SQS_LOW_PRIORITY_QUEUE = "${module.queue_low.id}"
   }
 
-  # TODO: Remove these once application is updated to use new vars.
-  legacy_config_vars = {
-    SQS_BACKFILL_QUEUE = "${module.queue_low.id}"
-    S3_KEY             = "${module.iam_user.config_vars["AWS_ACCESS_KEY"]}"
-    S3_SECRET          = "${module.iam_user.config_vars["AWS_SECRET_KEY"]}"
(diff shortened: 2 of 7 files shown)


DFurnes:
I've copied each environment's keys into the newly provisioned `dosomething-northstar-dev`, `dosomething-northstar-qa`, and `dosomething-northstar` buckets & confirmed everything looks good. I think that means we're in good shape to flip Northstar to the new S3, SQS, and IAM role.

DFurnes:
I've applied the changes in 22556fa and confirmed that everything works as expected.

DFurnes:
I'm currently syncing files in our old Rogue S3 buckets to their new counterparts:

```sh
aws s3 sync s3://ds-rogue-qa/uploads/reportback-items s3://dosomething-rogue-dev/uploads/reportback-items --profile ds # done!
aws s3 sync s3://ds-rogue-thor/uploads/reportback-items s3://dosomething-rogue-qa/uploads/reportback-items --profile ds # in progress…
aws s3 sync s3://ds-rogue-prod/uploads/reportback-items s3://dosomething-rogue/uploads/reportback-items --profile ds
```

Once that's done, I'll swap Rogue to use those new resources (as done for Northstar above).

DFurnes:
I've updated Northstar to use the standard `config_vars` from each Terraform module in DoSomething/northstar#823 and deployed to all environments, so removing those in 27b75bc.

DFurnes:
I've configured database migration tasks (in DMS, under `dosomething-replication-server` in the same `DoSomething` VPC and `default` security group as the RDS instances). Here's the mapping:

- [x] `rogue-db-dev`'s `rogue_qa` database → `dosomething-rogue-dev`'s `dosomething_rogue_dev`
- [ ] `rogue-db-dev`'s `rogue_thor` database → `dosomething-rogue-qa`'s `dosomething_rogue_qa`
- [ ] `rogue-db-prod`'s `rogue_prod` database → `dosomething-rogue`'s `dosomething_rogue`
- [ ] `phoenix-next`'s `phoenix_next` database → `dosomething-phoenix`'s `dosomething_phoenix`

DFurnes:
I updated the new databases' security group to the existing `mysql` group (which opens up public incoming traffic on 3306). In the future, I'd like to provision this security group with Terraform and pass it in as an argument.

I've also added a `force_public` setting to the S3 module in 2f0bd68...f5ebff4, which replicates the existing (non-ideal) behavior on these existing buckets. Once things are moved over & stabilized, we should absolutely revisit this and lock these down some more.

shared/s3_bucket/main.tf
```diff
@@ -13,6 +13,11 @@ variable "acl" {
   default     = "public-read"
 }
 
+variable "force_public" {
```
sheyd: Is this not recommended, since we don't want the S3 URI's to be directly accessible, except for from `rogue` domain serving them?

sheyd: 👹Totally agree. 💯

DFurnes:
Okay great! I've applied through af6ab27 and confirmed everything's working as expected. 🏁

DFurnes:
I finished one last "sync" to get the last couple reportbacks created in `ds-rogue-prod` into `dosomething-rogue` and updated the [weekly QA refresh](https://jenkins-thor.dosomething.org/job/Refresh%20Rogue-Thor%20S3%20from%20Prod/) job to use the new bucket names.

I deleted the old SQS queues, but let's keep the old S3 buckets around for now just in case.

DFurnes:
An interesting gotcha – it looks like DMS [doesn't migrate any indexes aside from primary keys](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Troubleshooting.html#CHAP_Troubleshooting.General.MissingSecondaryObjs), so we'll need to separately copy those over. That's frustrating. Some [resources](https://forums.aws.amazon.com/thread.jspa?threadID=266392) indicate that we can safely perform a normal dump and then use DMS to keep things up-to-date as we make the swap, so I'm going to test that out on `dosomething-rogue-qa` next and see how it goes.

DFurnes:
To keep things simpler, I'm also planning on just doing traditional dumps for `dosomething-phoenix-dev`, `dosomething-phoenix-qa`, and `dosomething-phoenix` (since there's no critical information here, aside from keeping sessions intact).

I also noticed that we've got both `dosomething-phoenix` and `dosomething-phoenix-preview` connected to the same database, which feels like a potential for a nasty surprise down the road. I'm going to create a fresh database for Preview so we don't have any chance for conflicts.

DFurnes:
Oh, and extra confusingly, we also have a JawsDB MariaDB Heroku add-on on that instance, but it's not getting used and the database is blank! I'm going to just remove that & swap Preview to use the fresh Terraform database that was provisioned for that app.

DFurnes:
I've copied over existing data into the new databases for `dosomething-phoenix-dev`, `dosomething-phoenix-qa`, and `dosomething-phoenix`, and I've created a fresh database for `dosomething-phoenix-preview` since it was previously reading and writing to production.

Going to swap these apps to use the new databases now (850d29e)!

DFurnes:
No issues with the `phoenix-next` database shut down, so created a final snapshot & terminated.

DFurnes:
Since @sheyd is out sick today, I'm going to merge this in and we can handle swapping over the Rogue databases in a follow-up branch tomorrow. I want to make sure we have good eyes on ensuring all the Quasar replication stuff still works with the new databases!

PR #100, opened 2019-01-08
Use new Rogue RDS instances.

DFurnes:
This pull request swaps Rogue to use the newly provisioned database instances. This is pending a little more work on getting a smooth database migration with DMS & ensuring that Quasar replication continues to work without a hitch.

### DIFF
--- a/applications/rogue/main.tf
+++ b/applications/rogue/main.tf
@@ -63,9 +63,8 @@ module "app" {
   web_size  = "${var.environment == "production" ? "Standard-2x" : "Standard-1x"}"
   web_scale = "${var.environment == "production" ? 2 : 1}"
 
-  # TODO: Add 'module.database.config_vars' once we've migrated
-  # records over to the new database instance.
   config_vars = "${merge(
+    module.database.config_vars,
     module.queue.config_vars,
     module.iam_user.config_vars,
     module.storage.config_vars,


DFurnes:
So we ran into one more gotcha with getting DMS set up... foreign key constraints! Since the `posts` and `signups` tables each have foreign key constraints applied to them, DMS is unable to truncate the table (and potentially would run into order-of-operations issues when inserting records too...)

```
Cannot truncate a table referenced in a foreign key constraint (`dosomething_rogue_qa`.`reactions`, CONSTRAINT `reactions_post_id_foreign` FOREIGN KEY (`post_id`) REFERENCES `dosomething_rogue_qa`.`posts` (`id`)) [1022502]
```

It seems like we've got two routes:

1. Continue working towards a zero-downtime migration with DMS by retrying the job with those two foreign key constraints removed. If that works, we can then re-add the constraints after the migration is finished.
2. Run a manual `mysqldump` and load into the new database. We just finished a test run of this on QA and it took 3.6 minutes to dump, but 15+ minutes to load that dump into the new instance. This may run a bit faster on production since we're transferring between two beefier instances, but we also have a moderately larger `events` table there (5.5GB instead of 4.3GB).

DFurnes:
If we decide to go the DMS route, here's what we'll have to do:

- [x] Dump schema from the existing database, and import into the new database.
- [x] Remove foreign key constraints from `reactions`, `posts`, and `reviews` tables.
- [ ] Start DMS job to migrate (ongoing) records to the new database; let it finish & validate.
- [ ] Place Rogue instance into maintenance mode, swap database config to point to the new instance.
- [ ] Re-add [foreign key constraints](
https://github.com/DoSomething/infrastructure/pull/100#issuecomment-452815180) to the new instance.
- [ ] Take the application out of maintenance mode.

I'll run through these steps on QA & see how things go. If all is well, we can tackle prod tomorrow.

DFurnes:
And since we only use Quasar's Rogue replication job for the `campaigns` table (which doesn't change frequently), I'd also like to hold on updating those parameter groups until _after_ the migration so that we don't add another variable into this equation!

DFurnes:
Alright, I've run through the [checklist](#issuecomment-452829330) on `dosomething-rogue-qa` and everything seems to have worked without issue! The initial load took a little over 2 hours (no downtime) and replication successfully copied over data until the moment I put the app into maintenance mode.

The downtime (swapping credentials, finishing replication task, and re-adding foreign keys) lasted just a little over six minutes, and we seem to have had no data loss or other inconsistencies.

DFurnes:
Here's the checklist for running this migration & swapover on production:

- [x] Dump schema from the existing database, and import into the new database.
- [ ] Remove foreign key constraints from `reactions`, `posts`, and `reviews` tables.
- [ ] Start DMS job to migrate (ongoing) records to the new database; let it finish & validate.
- [ ] Place Heroku app into maintenance mode, swap database config to point to the new RDS instance.
- [ ] Stop the DMS job & wait for it to clean up its work. (~5 minutes)
- [ ] Re-add [foreign key constraints](
https://github.com/DoSomething/infrastructure/pull/100#issuecomment-452815180) to the new instance. (~1 minute)
- [ ] Take the Heroku app out of maintenance mode.

DFurnes:
Just rebased to grab the latest changes from `master`.

DFurnes:
Applied 8e2053e, which at this point was a no-op! I'm going to merge this in & finish cleanup tasks.

sheyd:
PR 100!

PR #101, opened 2019-01-08
Update legacy SQS config vars.

DFurnes:
I'd accidentally set the wrong "legacy" SQS environment variables for Rogue in #99, so we've been continuing to use the same old IAM role for SQS (although with the new queue). This pull request updates these to be [what Rogue currently uses](https://github.com/DoSomething/rogue/blob/e767fe012d7318c44716e0fda37e6faafa7929db/config/queue.php#L51-L58) so we can safely delete these old IAM users.

### DIFF
--- a/applications/rogue/main.tf
+++ b/applications/rogue/main.tf
@@ -45,10 +45,10 @@ variable "with_newrelic" {
 locals {
   # TODO: Remove these once application is updated to use new vars.
   legacy_config_vars = {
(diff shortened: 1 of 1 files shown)


DFurnes:
I'll follow up with a Rogue pull request to use the new standardized vars (a la DoSomething/northstar#823) later today.

DFurnes:
Applied ba76a17 & everything seems a-okay. Will confirm that those old IAM roles are now unused later this afternoon and delete them if so.

PR #103, opened 2019-01-09
Removes the need for serving static pages from Ashes (Dev/QA).

mirie:
There are no remaining static pages that should be served from Ashes.

Is this the right way to do this @DFurnes?


DFurnes:
I've applied 585903c. Going to merge this in!

PR #104, opened 2019-01-09
Removes the need to serve static pages from Ashes (Prod).

mirie:
Looked good on QA: 
- https://qa.dosomething.org/us/about/its-your-world
- as well as a previously moved one: https://qa.dosomething.org/us/about/our-team

There are no static pages remaining that should be served from Ashes! It's Your World is the last page on Production :)

Also,  It's Your World Page on master : https://app.contentful.com/spaces/81iqaqpfd8fy/entries/4LaK1Vc1mwgcm8e8Gi8w6k is published

### DIFF
--- a/dosomething/fastly-frontend/ashes_init.vcl
+++ b/dosomething/fastly-frontend/ashes_init.vcl
@@ -304,10 +304,3 @@ table phoenix_facts {
   "11-facts-about-birth-control": "true",
   # ...
 }
-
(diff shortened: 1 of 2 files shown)


DFurnes:
Applied 0269556 & confirmed the [It's Your World](https://www.dosomething.org/us/about/its-your-world) page serves from Phoenix now!

PR #107, opened 2019-01-14
Add application database user.

DFurnes:
This pull request adds application users to each of our MariaDB databases, so that applications don't have [full "root" access](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.MasterAccounts.html) to the RDS instance (say to add/remove other users or set up replication).

After provisioning these new limited users, I'll test them out on a few dev & QA instances before setting them in the module's outputs so they take effect across the board.

### DIFF
--- a/shared/mariadb_instance/main.tf
+++ b/shared/mariadb_instance/main.tf
@@ -61,10 +61,6 @@ data "aws_ssm_parameter" "database_username" {
   name = "/${var.name}/rds/username"
 }
 
-data "aws_ssm_parameter" "database_password" {
-  name = "/${var.name}/rds/password"
-}
-
 locals {
   # Should backups be enabled, and if so for how long? We keep backups for
   # the maximum window (30 days) on production, and otherwise the minimum
@@ -96,6 +92,13 @@ resource "aws_db_parameter_group" "replication_settings" {
   }
 }
 
+resource "random_string" "master_password" {
+  length = 32
+
+  # We can't use '@' or '$' in MySQL passwords.
+  override_special = "!#%&*()-_=+[]{}<>:?"
+}
+
 resource "aws_db_instance" "database" {
   identifier = "${var.name}"
   name       = "${local.safe_name}"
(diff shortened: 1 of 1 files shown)


DFurnes:
I'm going to apply 5c565de + 95f233e to test these new users. This will _not_ modify any existing resources (applications will continue to use the master RDS user as they always have).

DFurnes:
> One question regarding perms, I’m assuming answer it yes, do we want to give app DB user ability to drop tables?

Good thought! I think we want the answer to this to be "yes", since we'll sometimes need it for migrations to work successfully (and I feel more comfortable letting a peer-reviewed migration do something like that than having an "operator" manually drop the table).

I've updated the config to be explicit in what permissions are granted in 73bdc99.

DFurnes:
I'm going to manually test these new credentials on `dosomething-rogue-dev` and `dosomething-rogue-qa` before rolling them out to all of our applications. Stay tuned!

DFurnes:
Confirmed that everything seems to work correctly (viewing galleries, creating signups/posts, and reviewing), and ran a successful refresh (dropped all tables & re-ran migrations) on development.

DFurnes:
Since everything seems good, I think we're ready to apply this across the board in a9eec0a!

DFurnes:
Applied, and confirming that everything looks good across the board.

DFurnes:
Finally, I'm going to re-generate the master passwords for each of these instances in a28d6f7. According to [AWS's documentation](https://docs.aws.amazon.com/cli/latest/reference/rds/modify-db-instance.html), this will not incur an outage or require rebuilding the instance:

> Changing this parameter doesn't result in an outage and the change is asynchronously applied as soon as possible. Between the time of the request and the completion of the request, the MasterUserPassword element exists in the PendingModifiedValues element of the operation response.

DFurnes:
Applied that change, and again confirming everything is still okay!👌

DFurnes:
Confirmed that old master credentials are invalid, and removed those parameters from SSM.

PR #111, opened 2019-01-17
Standardize GraphQL instances with app module.

DFurnes:
This pull request updates our GraphQL instances to use a standard `applications/graphql` module.

As part of this, I've renamed the `laravel_app` module to `heroku_app` and added a `stack` argument (that, for now, accepts either `laravel` or `express`) since 90% of this module is stack-agnostic.


shared/heroku_app/main.tf
```diff
@@ -78,34 +83,61 @@ data "aws_ssm_parameter" "slack_deploy_webhook" {
...
+      # Configure logging for Heroku drain:
+      APP_LOG = "errorlog"
+
+      # Alongside the Trusted Proxy module, this allows us to use SSL behind
+      # Heroku's load balancer. Heroku strips these headers on incoming traffic
+      # so it's safe to trust all (and we can't know their specific IPs).
+      TRUSTED_PROXY_IP_ADDRESSES = "**"
+
+      # If we're using Redis for this app, tell Laravel to use that for cache & session store.
+      CACHE_DRIVER   = "${var.with_redis ? "redis" : "file"}"
```
sheyd: Are there no instances by default where we'll use caching for express apps? Pretty sure *no* since this is to setup GraphQL, but it's only notable diff I've noticed, so asking, just in case. :)

shared/heroku_app/main.tf
```diff
@@ -11,6 +11,11 @@ variable "name" {
   description = "The application name."
 }
 
+variable "stack" {
```
sheyd: Stack, framework, 🤷‍♂️ Naming is hard.

shared/heroku_app/main.tf
DFurnes: We actually do use Redis caching in most (or all?) of our Express apps, but it's not configured with a swappable driver. This `CACHE_DRIVER` is a Laravel-specific setting.

shared/heroku_app/main.tf
DFurnes: Framework could also make sense! I don't have a strong preference either way.

DFurnes:
Applied & confirmed everything is still groovy.

shared/heroku_app/main.tf
DFurnes: Renamed to `framework` in efb5ff4!

sheyd: 🐦

PR #116, opened 2019-01-18
Set 'ENGINE_API_KEY' to Apollo Engine API key.

DFurnes:
The method for configuring Apollo Engine [changed](https://www.apollographql.com/docs/apollo-server/migration-engine.html) in Apollo Server 2.x, and so we'll want to provide this API key under the `ENGINE_API_KEY` environment variable instead. Refs DoSomething/graphql#41.


DFurnes:
Applied! Merging.

Issue #98, opened 2018-12-21
Retire celebsgonegood.com infrastructure


mshmsh5000:
This is now being served by Instapage, so we can truly retire everything related to CGG, including the Fastly property.

DFurnes:
We're still serving a static snapshot of http://www.athletesgonegood.com off of this Fastly property & S3 bucket, but definitely worth cleaning up the CGG part! 💯

mshmsh5000:
So I imagine for an issue like this, we'd first delete the references from the [Terraform config](https://github.com/DoSomething/infrastructure/blob/master/voting-app/main.tf#L5), and then optionally create a follow-up issue to delete the underlying resource. Is that right?

In this case, we might want to keep the underlying S3 bucket for posterity, so just want to clarify the process.

DFurnes:
Yep, we'd delete the references from the Terraform config and those underlying resources would be removed when we ran `terraform apply`. For things we want to keep, we should keep the Terraform resource.

mshmsh5000:
Ah, so maybe we should just leave these be? Or delete the Fastly references but keep the S3 reference?

DFurnes:
I think it depends how likely we are to want to reuse that bucket. Since it's such an old snapshot of the site (from 2016), I'd imagine its unlikely we'll want it back. And worst case, I had also placed those snapshots in [this repository](https://github.com/DoSomething/voting-app-static) if we need to re-create.

DFurnes:
This was done in #117! 🎬

PR #117, opened 2019-01-22
Remove CGG from Voting App config.

DFurnes:
Cleaning this up so it can be a concise example for tomorrow's roundtable.

References #98.


DFurnes:
I applied this so I could grab a few screenshots. 🤠 Gonna merge it in!

PR #119, opened 2019-01-23
Experimental/WIP: Manage Lambda functions with Terraform.

DFurnes:
**This is a work-in-progress experimental branch.** I'm hoping to be able to manage AWS Lambda functions with Terraform (rather than CloudFormation as we've previously done), so that we can easily visualize & manage them alongside all of our other infrastructure.

Roughly following this [Terraform guide](https://learn.hashicorp.com/terraform/aws/lambda-api-gateway).


applications/graphql-lambda/main.tf
```diff
@@ -0,0 +1,72 @@
...
+  bucket = "${aws_s3_bucket.deploy.id}"
+  key    = "${local.safe_name}.zip"
+  source = "${file("${path.module}/example.zip")}"
+}
+
+resource "aws_lambda_function" "function" {
+  function_name = "${var.name}"
+  handler       = "main.handler"
+
+  # The bucket name as created earlier with "aws s3api create-bucket"
```
mshmsh5000: @DFurnes So, in this example, you would first need the Lambda build packaged up locally as a zip?

DFurnes: Ah, that comment is a lie! My hope is that we can push up a "default" function (stored in `example.zip`) when provisioning the new resource and then manage uploading new releases via a separate IAM "deploy" role without invalidating state.

DFurnes:
I've applied this first iteration, and confirmed we can trigger the function via the `aws` CLI:

```
aws lambda invoke --region=us-east-1 --function-name=dosomething-graphql-dev /dev/stdout
{"statusCode":200,"headers":{"Content-Type":"text/html; charset=utf-8"},"body":"<p>Hello world!</p>"}
```

DFurnes:
And now we've got a deploy user w/ limited permissions to this app's deploy bucket & function code:

```
aws s3 cp example_v2.zip s3://dosomething-graphql-dev-deploy/dosomething_graphql_dev.zip
aws lambda update-function-code --function-name dosomething-graphql-dev --s3-bucket dosomething-graphql-dev-deploy --s3-key dosomething_graphql_dev.zip --publish --profile graphql-deploy

aws lambda invoke --region=us-east-1 --function-name=dosomething-graphql-dev --profile ds /dev/stdout
{"statusCode":200,"headers":{"Content-Type":"text/html; charset=utf-8"},"body":"<p>Yo, v2!</p>"}
```

And (phew), re-running a `terraform plan` after making this change does _not_ see this as "dirty".

DFurnes:
Alright, this is ready for some load testing! This pull request introduces a `lambda_function` module that configures a deploy bucket (seeded with a "hello world" example app), Lambda function, CloudWatch logs, and separate execution & deploy lambda roles.

I've also swapped the cache backend from Redis to DynamoDB (using a swappable cache driver added in DoSomething/graphql#46). This is handy for two reasons:
1. We don't need to move this Lambda function into a VPC (required for ElastiCache usage), which adds complexity (since we need to provision or choose subnets & security groups).
2. We only pay for as much as we use, using DynamoDB's on-demand pricing model.

We can then use the following snippet to deploy this function (using the provisioned deploy key):

```
zip -r - . | aws s3 cp - s3://dosomething-graphql-dev-deploy/release.zip --profile dosomething-graphql-dev-deploy && aws lambda update-function-code --function-name dosomething-graphql-dev --s3-bucket dosomething-graphql-dev-deploy --s3-key release.zip --publish --profile dosomething-graphql-dev-deploy
```

We can obviously simplify that a bit (and configure it to run automatically with CI), but it's nice to see everything works as expected! And unlike Serverless, we have no fear that a deploy will unexpectedly create or destroy production infrastructure (or even could).

DFurnes:
Lambda deployment for **dosomething-graphql-dev**:

https://8ag15810zh.execute-api.us-east-1.amazonaws.com/development/

DFurnes:
To start, I ran a quick load test on the `topic` query (since that resolver caches downstream API requests, and so is a nice way to test the GraphQL server in isolation from other systems):

<details>
<summary><strong>testing 40 concurrent users against the Heroku backend</strong></summary>

```
siege -c40 -t60S --content-type "application/json" 'https://graphql-dev.dosomething.org/graphql POST {"query":"{topic(id: \"G8LCAYO9WuGC2UW8mQi6w\") { name,  ...on AskYesNoBroadcastTopic {text,saidYes}}}"}'

Lifting the server siege...
Transactions:		        3384 hits
Availability:		      100.00 %
Elapsed time:		       59.03 secs
Data transferred:	        2.66 MB
Response time:		        0.69 secs
Transaction rate:	       57.33 trans/sec
Throughput:		        0.05 MB/sec
Concurrency:		       39.46
Successful transactions:        3384
Failed transactions:	           0
Longest transaction:	        6.83
Shortest transaction:	        0.12
```

</details>

<details>
<summary><strong>testing 40 concurrent users against the Lambda backend</strong></summary>

```
siege -c40 -t60S --content-type "application/json" 'https://8ag15810zh.execute-api.us-east-1.amazonaws.com/development/ POST {"query":"{topic(id: \"G8LCAYO9WuGC2UW8mQi6w\") { name,  ...on AskYesNoBroadcastTopic {text,saidYes}}}"}'

Lifting the server siege...
Transactions:		         772 hits
Availability:		       92.23 %
Elapsed time:		       59.50 secs
Data transferred:	        0.61 MB
Response time:		        2.82 secs
Transaction rate:	       12.97 trans/sec
Throughput:		        0.01 MB/sec
Concurrency:		       36.59
Successful transactions:         772
Failed transactions:	          65
Longest transaction:	       29.68
Shortest transaction:	        0.21
```
</details>
<br />

Well, that's funny – it mostly performed admirably, but we got 65 transactions that timed out and tanked the average response time. These stragglers are each taking around 29 seconds, way above our 15 second Lambda execution limit. (It turns out this is the [default API Gateway integration timeout](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html). We can & should probably reduce this.)

When I tried another query that doesn't hit the DynamoDB cache (in this case, just asking about the schema with `{__schema {queryType {name}}}`), I don't see any timeout issues – even when doubling the load-test concurrency to 80 users! So that seems to narrow down the culprit.

I'm continuing to dig into why DynamoDB requests may be sporadically timing out.

DFurnes:
I was able to dig up some references to timeouts with Lambda & DynamoDB, but the suggested fixes are all for an outdated TLS issue (See [1](
https://stackoverflow.com/questions/37466847/dynamodb-causing-lambda-timeout), [2](
https://stackoverflow.com/questions/34411314/aws-dynamodb-gives-write-eproto), [3](
https://aws.amazon.com/premiumsupport/knowledge-center/write-eproto-error/)). Some others ([4](
https://github.com/aws/aws-sdk-js/issues/1086)) suggested VPC config issues (but we're not running this in a VPC) or memory woes (bumping to 512MB had no effect).

I enabled CloudWatch logs on our API Gateway account to confirm there aren't any obvious errors being logged there & used [X-Ray](https://aws.amazon.com/xray/) to peek into some traces for this function. Most look like you'd expect:

screen shot 2019-01-25 at 5 20 41 pm

However, a subset of requests time out like this:

screen shot 2019-01-25 at 5 21 09 pm

…or like this:

screen shot 2019-01-25 at 5 53 05 pm

This suggests our function never runs at all in these cases, which is backed up by the service map:

screen shot 2019-01-25 at 5 31 31 pm

…anyways, that's as far as I've got! Tune back in next week. 📺 🙃

DFurnes:
Jumping back in to this. Re-capping and stepping through things step-by-step:

- I'm seeing 502/504 timeouts for roughly 25% of requests under load.
- This happens for _both_ requests that hit DynamoDB and ones that don't.
- It does _not_ seem to happen when I load test the "hello world" example application that simply returns a success response. That makes me think that it's some sort of issue in our code, even though the above traces seemed to suggest the request was never hitting the Lambda function.
- Likewise, it does _not_ happen when load-testing [an example Apollo Server](https://www.apollographql.com/docs/apollo-server/deployment/lambda.html) with the example schema and simple `{ hello }` query.
- As soon as I un-comment [the `esm` hook](https://github.com/DoSomething/graphql/blob/90ff160c325b0b13fd3cb06c9c15d0ef7bb879b9/main.js#L5) in our Lambda entry point, I see the issue start to appear (even _without_ requiring any of our ES modules). That's not what I expected at all!

DFurnes:
Re-based against `master` to get latest merged changes.

sheyd:
🏂 👢

DFurnes:
I was finally able to track this down to Lambda's "initialization" stage, by increasing the function's timeout far beyond API Gateway's 29 second timeout. This shows that the "actual" function completes in a fairly reasonable 3.1 seconds, but since it is stuck waiting for so long it's already missed the "response" window by the time it's even started. Gah!

screen shot 2019-01-28 at 5 05 06 pm

I've [reached out to Amazon's support](https://console.aws.amazon.com/support/cases#/5734584801/en) to see if they can provide any suggestions. 🤞

applications/graphql-lambda/dynamodb-policy.json.tpl
```diff
@@ -0,0 +1,34 @@
+{
```
sheyd: Assuming this is the standard copied from AWS grants permissions groups console?

applications/graphql-lambda/main.tf
```diff
@@ -0,0 +1,123 @@
...
+  name = "/${var.name}-lambda/apollo/api-key"
+}
+
+resource "random_string" "app_secret" {
+  length = 32
+}
+
+locals {
+  # Environment (e.g. 'dev' or 'DEV')
+  env = "${var.environment == "development" ? "dev" : var.environment}"
```
sheyd: Why might we have different `dev` vs `development`? Is this local vs Heroku hosted?

main.tf
```diff
@@ -60,7 +60,7 @@ provider "heroku" {
 # some legacy servers on EC2. AWS credentials are stored
 # using the `aws` CLI (see installation instructions).
 provider "aws" {
-  version = "~> 1.41"
+  version = "~> 1.56"
```
sheyd: Yeah, UPGRADE!

shared/lambda_function/main.tf
```diff
@@ -0,0 +1,194 @@
...
+  description = "The application name."
+}
+
+# Optional variables:
+variable "config_vars" {
+  description = "Environment variables for this application."
+  default     = {}
+}
+
+locals {
```
sheyd: This is cool, just looked up what `locals` does.

shared/lambda_function/main.tf
```diff
@@ -0,0 +1,194 @@
...
+}
+
+resource "aws_api_gateway_resource" "proxy" {
+  rest_api_id = "${aws_api_gateway_rest_api.gateway.id}"
+  parent_id   = "${aws_api_gateway_rest_api.gateway.root_resource_id}"
+  path_part   = "{proxy+}"
+}
+
+# We need separate proxies for the "root" path '/' and everything else '/*'
+# because reasons. Frustrating, but luckily we can abstract this away.
```
sheyd: roll eyes

applications/graphql-lambda/main.tf
DFurnes: The GraphQL app expects this environment variable to be provided as `dev`. We store environment as the full string `development` everywhere else in Terraform.

We should reconcile these so it's consistent everywhere!

applications/graphql-lambda/dynamodb-policy.json.tpl
DFurnes: [Yup](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_dynamodb_specific-table.html).

applications/graphql-lambda/main.tf
DFurnes: > Is this local vs Heroku hosted?

No, local apps would have `APP_ENV=local`. Our [development environment](http://docs.dosomething.org/engineering/overview#environments) is what developers connect their locals to when working on code (since we don't have a convenient way to spin up & maintain an entire local environment for folks, e.g. with Docker Compose)

PR #120, opened 2019-01-28
Switch to serving search endpoint from Phoenix.

weerd:
This PR just switches serving the search endpoint from Phoenix instead of Ashes so that we can activate the new search bar in Phoenix!


DFurnes:
Applied! 🔍

PR #121, opened 2019-01-28
Adding Fivetran user to MariaDB instances for Quasar replication.

sheyd:
Adding dedicated `fivetran` user to MariaDB instances so we can replicate data from Rogue QA/Prod to Quasar QA/Prod.

@DFurnes I've run the `make plan`, and the out put looks sane (as per perusing for `fivetran` related changes), but would love a 👬 for this one, since it's adding/altering DB credentials, and our state is a bit in flux with the GraphQL Lambda experiments.

### DIFF
--- a/shared/mariadb_instance/main.tf
+++ b/shared/mariadb_instance/main.tf
@@ -189,6 +189,28 @@ resource "mysql_grant" "readonly" {
   privileges = ["SELECT"]
 }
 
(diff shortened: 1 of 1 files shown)


DFurnes:
Fixed an issue with grant permissions in 75356d9 & then successfully applied! Merging.

PR #131, opened 2019-02-04
Restructuring legacy redirects

weerd:
This PR updates the `projectpages` snippets, renaming them to `legacy_redirects`, and includes a few more redirects.

All legacy related redirects will be `8xx` errors, with `800` being a redirect to the homepage.

This also updates some redirects which have their status set to `302` to be `301` instead, based on some IRL convos with @DFurnes @ngjo and @mshmsh5000 🤝

### DIFF
--- a/dosomething-dev/fastly-frontend/legacy_redirects_recv.vcl
+++ b/dosomething-dev/fastly-frontend/legacy_redirects_recv.vcl
@@ -1,14 +1,14 @@
 # Redirect */actnow/* page requests.
-if (req.url ~ "(?i)^\/((us|mx|br)\/)?actnow\/") { error 800 "Not Found"; }
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?actnow\/") { error 800 "/us"; }
 
 # Redirect */programs/* page requests.
-if (req.url ~ "(?i)^\/((us|mx|br)\/)?programs\/") { error 800 "Not Found"; }
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?programs\/") { error 800 "/us"; }
   
 # Redirect */project/* page requests.
-if (req.url ~ "(?i)^\/((us|mx|br)\/)?project(s)?\/") { error 800 "Not Found"; }
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?project(s)?\/") { error 800 "/us"; }
 
 # Redirect */about/press/* page requests.
-if (req.url ~ "(?i)^\/((us|mx|br)\/)?about\/press\/") { error 810 "Not Found"; }
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?about\/press\/") { error 800 "/us/about/our-press"; }
 
 # Redirect */about/team/* page requests.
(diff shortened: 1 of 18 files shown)


dosomething-dev/fastly-frontend/legacy_redirects_recv.vcl
```diff
@@ -0,0 +1,14 @@
+# Redirect */actnow/* page requests.
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?actnow\/") { error 800 "Not Found"; }
```
DFurnes: Instead of three separate error codes, what if we shared `800` for all of these and passed a parameter? I haven't tested this out before, but I think it might work...

```vcl
if (req.url ~ "(?i)^\/((us|mx|br)\/)?actnow\/") { error 800 "/us"; }
```

```
if (obj.status == 800) {
  set obj.http.Location = obj.response;
  set obj.status = 301;
  return (deliver);
}
```

weerd: 🤔 yeah, potentially. Does having the separate codes help in terms of knowing what's getting requested a lot? If not I guess that could make it easier?

weerd: @DFurnes ended up just going for it and trying this out. Ultimately it'll make adding more legacy redirects easier in the future! Lemme know if all looks good and works in a dry-run 😉

DFurnes:
Applied this without error. Now just running through some testing!

dosomething/fastly-frontend/legacy_redirects_recv.vcl
```diff
@@ -0,0 +1,14 @@
...
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?programs\/") { error 800 "/us"; }
+  
+# Redirect */project/* page requests.
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?project(s)?\/") { error 800 "/us"; }
+
+# Redirect */about/press/* page requests.
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?about\/press\/") { error 800 "/us/about/our-press"; }
+
+# Redirect */about/team/* page requests.
+if (req.url ~ "(?i)^\/((us|mx|br)\/)?about\/team\/") { error 800 "/us/about/our-people"; }
```
DFurnes: One potentially small gotcha: since this regex ends in the `/`, this won't match on [`/about/team`](https://www.dosomething.org/us/about/team).

Are we getting any hits there?

weerd: Not that I saw, but also don't want to target for not ending in `/` since we may have some page in the future that begins with team and is hyphenated. We can just do an Aurora redirect for this one.

DFurnes: Oh sure, Aurora redirect works for me! 👍

weerd: Added the redirect to Aurora to help :)

PR #132, opened 2019-02-05
Terraform/Lambda clean-up.

DFurnes:
This pull request includes some clean-up to make the Terraform/Lambda prototype more re-useable:

🌐 Extracts `api_gateway_proxy` module so we can provision HTTP-less Lambda functions.

💰 Extracts `dyanamodb_cache` module so we could re-use this same cache setup elsewhere.

🍃 Removes some environment variables that aren't used by the Lambda GraphQL app.


DFurnes:
I've somehow made a boo-boo, since the API Gateway permission is no longer being applied...

```
Execution failed due to configuration error: Invalid permissions on Lambda function
```

👀

DFurnes:
I had been hoping to be able to map API gateway paths to functions with a dictionary, but that seems difficult with the way that API Gateway resources work alongside the "root" resource. For now, I've just renamed this module into a more purpose-built `api_gateway_proxy`.

DFurnes:
I extracted the DynamoDB cache into a re-usable module in abbae8e and 7916c81.

Issue #11, opened 2018-10-05
Add DNS records to Terraform.


DFurnes:
We should add our DNS records to Terraform so we can manage subdomains & hook them up to the right backend in code. I could've sworn DNSMadeEasy was a third-party provider last time I looked, but it turns out it's [included out-of-the-box](https://www.terraform.io/docs/providers/dme/index.html).

DFurnes:
Or not! The Terraform DNSMadeEasy provider [has been archived](https://github.com/terraform-providers/terraform-provider-dme).

mshmsh5000:
Oh snaps! This is an example of why I want a network of Terraform practitioners to lean on.

DFurnes:
Closing in favor of [#165477075](https://www.pivotaltracker.com/story/show/165477075).

PR #133, opened 2019-02-05
Add custom domain support to API Gateway module.

DFurnes:
This pull request adds support for attaching a custom domain to an API Gateway stage. For now, I'm just importing our existing wildcard certificate using the [`aws_acm_certificate`](`https://www.terraform.io/docs/providers/aws/d/acm_certificate.html) data resource (since we don't use Route53 and DNSMadeEasy's provider is [kaput](https://github.com/DoSomething/infrastructure/issues/11#issuecomment-460815237)).

### DIFF
--- a/applications/graphql-lambda/main.tf
+++ b/applications/graphql-lambda/main.tf
@@ -8,8 +8,9 @@ variable "name" {
   description = "The application name."
 }
 
-variable "papertrail_destination" {
-  description = "The Papertrail log destination for this application."
+variable "domain" {
+  description = "The domain this application will be accessible at, e.g. graphql-lambda.dosomething.org"
+  default     = ""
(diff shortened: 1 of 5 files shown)


DFurnes:
Applied successfully, and confirmed we've got a custom domain attached to the `graphql-lambda-dev` API Gateway! We attached the provided CNAME to `graphql-lambda-dev.dosomething.org` in DNSMadeEasy (Sena: "the worst") and are just waiting for the cert to finish provisioning. ⏳

DFurnes:
Hawt: https://graphql-lambda-dev.dosomething.org 🔥

DFurnes:
Added custom domains for QA & production in 340ec1c.

PR #134, opened 2019-02-06
Forward logs to Papertrail for Terraform-provisioned Lambdas.

DFurnes:
**Work in progress.** This pull request uses our [lambda-papertrail](https://github.com/dosomething/lambda-papertrail) function to forward logs for our new Terraform-provisioned Lambdas from CloudWatch to Papertrail. The first step is to provision the functions, then configure CI for deploys, and finally attach them to the appropriate log groups.

### DIFF
--- a/applications/graphql-lambda/main.tf
+++ b/applications/graphql-lambda/main.tf
@@ -1,5 +1,6 @@
 # Experimental: This module builds a serverless GraphQL instance.
 
+# Required variables:
 variable "environment" {
   description = "The environment for this application: development, qa, or production."
 }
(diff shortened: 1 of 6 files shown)


DFurnes:
It's worth noting that it's absolutely not necessary to re-provision this application with Terraform in order to use it to forward logs for these apps (we could reference the existing function from the [CloudFormation stack outputs](https://www.terraform.io/docs/providers/aws/r/cloudformation_stack.html)). However, I think it's nice to have a consistent deployment workflow & it's a good test-case for using our template with a different type of Lambda. 😺

DFurnes:
Ran into an issue when trying to attach the log subscription in 01d664a:

> aws_cloudwatch_log_subscription_filter.papertrail_subscription: InvalidParameterException: Could not execute the lambda function. Make sure you have given CloudWatch Logs permission to execute your function.

Sorted out by attaching to the log group ARN (instead of to the destination Lambda), and confirmed that the "trigger" is listed in dashboard. Finally, confirmed that the logs are [showing up in Papertrail](https://papertrailapp.com/systems/dosomething-graphql-dev-lambda/events).

mshmsh5000:
A thing of beauty is a joy forever. 👍

PR #136, opened 2019-02-06
Add Lookerbot resources.

mshmsh5000:
These changes declare an S3 bucket for our [Slack Lookerbot](https://github.com/looker/lookerbot) deploy, so it can publish visualizations to S3 and share those in Slack.

### DIFF
--- a/applications/lookerbot/main.tf
+++ b/applications/lookerbot/main.tf
@@ -1,0 +1,30 @@
+# This template builds an S3 bucket for a Lookerbot instance.
+#
+# See https://github.com/looker/lookerbot#amazon-s3
+#
+# Manual setup steps:
+#   - Finally, head to the S3 Bucket in AWS after provisioning, and go to the
+#     Permissions -> Public Access Settings screen. Check all the options.
+#
+# NOTE: We'll move more of these steps into Terraform over time!
+
+# Required variables:
+variable "name" {
+  description = "The application name."
+}
+
+module "iam_user" {
+  source = "../../shared/iam_app_user"
+  name = "${var.name}"
+}
+
+module "storage" {
+  source = "../../shared/s3_bucket"
+
+  name = "${var.name}"
+  user = "${module.iam_user.name}"
+}
+
+output "name" {
+  value = "${var.name}"
+}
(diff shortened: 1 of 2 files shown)


applications/lookerbot/main.tf
```diff
@@ -0,0 +1,30 @@
+# This template builds an S3 bucket for a Lookerbot instance.
+#
+# See https://github.com/looker/lookerbot#amazon-s3
+#
+# Manual setup steps:
+#   - Finally, head to the S3 Bucket in AWS after provisioning, and go to the
```
DFurnes: It'd be nice to outline the manual steps for setting this up here, e.g:

```
- Head to <https://git.io/fh9DS> and hit "Deploy to Heroku" to provision the app.
- Set the required environment variables via Heroku's admin panel. <https://git.io/fh9Dy>
- Set the `SLACKBOT_S3_BUCKET`, `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY`
  environment variables from the resources we've provisioned in this module.
```

We also probably don't want to lock down public access here, since this is used for file hosting.

mshmsh5000: Ah, good point. We can remove that S3 instruction entirely, can't we?

DFurnes: Yep!

mshmsh5000:
Ohhhhhhh.

```
* module.quasar.module.lookerbot.module.storage.aws_s3_bucket.bucket: 1 error(s) occurred:

* aws_s3_bucket.bucket: Error creating S3 bucket: BucketAlreadyExists: The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again.
	status code: 409, request id: 552ED495B970EBA1, host id: dcAKrjWxno9tlLouurDmKMtkZMGaMtqzgO1YTA+JGFcTfh1JGiKqvz5tqqWi1sZgZbJCxGeFoYE=
```

DFurnes:
Ahhh! That's how they get ya. How about we prefix with `dosomething-`?

mshmsh5000:
Resolved in #142 / #144.

PR #140, opened 2019-02-07
Serverless demo!

DFurnes:
👋


DFurnes:
I sneakily added the custom domain (4cc01c4) and fixed a bug with attaching more than one log group to the Papertrail forwarder function (add855f). Merging! 🔨

Issue #129, opened 2019-02-01
Experiment: Serverless GraphQL, with Terraform & CircleCI


DFurnes:
We use GraphQL to power our submission & campaign galleries, and our new user profiles. Starting [this quarter](https://docs.google.com/document/d/1Z0LesRih2rql2mumuM3126kkZC2Qw4IbCHMAl7iw77A/edit), we're also using GraphQL to fetch data for broadcasts in Gambit. As we send more traffic to GraphQL, we want to make sure things keep running smoothly!

One of the difficult things about scaling DoSomething.org is that our traffic is not consistent. For example, here's what traffic to [`graphql.dosomething.org`](https://graphql.dosomething.org) has looked like over the past 3 days:

![production-rpm-last-3-days](https://user-images.githubusercontent.com/583202/52079553-d827fc80-2563-11e9-913b-bfb898186d65.png)

Rather than have systems over-provisioned during low-traffic periods (or require manual intervention to scale up and down), I wanted to explore using "serverless" techologies – Lambda and DynamoDB – to scale up and down transparently and only pay for what we use.

I also wanted to take the opportunity to see whether we could manage our serverless applications with Terraform, since that's where we've begun collecting the rest of our infrastructure. Our previous foray into this world, [Bertly](https://github.com/DoSomething/bertly), was built with the [Serverless framework](https://serverless.com) which provides a batteries-included starting point but locks us into provisioning via per-app CloudFormation templates.

There are a few downsides to this forced dependency on CloudFormation:
1. CloudFormation is only able to provision AWS resources, so if we wanted to depend on a resource not built by Amazon, it'd have to be provisioned and hooked up by hand.
2. We lose the ability to connect to other Terraform resources automatically (via module inputs and outputs), and therefore increase the chance of accidentally connecting the "wrong wires" together.
3. Because Serverless provisions app infrastructure as part of the deploy process, it means any developer (or CI service) that's able to deploy a serverless application would also have read & write access to a large portion of our AWS resources. Yikes!
4. It adds conceptual overhead – instead of having everything in one place, it's one more tool to learn.

## Goal
With that in mind, I set out to move our [GraphQL service](https://graphql.dosomething.org) to run on "vanilla" Lambda, using Terraform to provision the deployment bucket, function, roles, logging, and API gateway.

This work was mostly tracked in #119, DoSomething/graphql#46, and #128. 

## Process
Here's a high-level overview of the work that went into this:

[...]

PR #146, opened 2019-02-11
Add support for API Gateways with named routes.

DFurnes:
The last bit of cleanup for #129! This pull request adds support for using API Gateway with "named" routes pointing to different Lambdas (rather than only as a proxy). I've kept the existing `api_gateway_proxy` module as a "shortcut" for cases where it makes sense.

This pull request also removes the `var.environment` argument from each of these API Gateway modules, and instead names the (single) stage as `default`. (Rather than using stages, we keep development & QA applications in their own gateways for isolation & independent testing).

**Module documentation:** [API Gateway](https://github.com/DoSomething/infrastructure/tree/api-gateway-nonproxy/components/api_gateway), [API Gateway (Proxy)](https://github.com/DoSomething/infrastructure/tree/api-gateway-nonproxy/components/api_gateway_proxy)

### DIFF
--- a/applications/graphql-lambda/main.tf
+++ b/applications/graphql-lambda/main.tf
@@ -80,7 +80,6 @@ module "gateway" {
   source = "../../components/api_gateway_proxy"
 
   name                = "${var.name}"
-  environment         = "${var.environment}"
   function_arn        = "${module.app.arn}"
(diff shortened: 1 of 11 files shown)


DFurnes:
Applied these changes & all seems good (just waiting on custom domains to provision).

PR #149, opened 2019-02-11
Swap GraphQL Development & QA to the new Lambda backend.

DFurnes:
This pull request (should, we'll see) attach `graphql-dev.dosomething.org` and `graphql-qa.dosomething.org` to the appropriate API Gateways (and then once that's all provisioned behind the scenes, we can swap the DNS target for those domains for a zero-downtime switch).

Once these are swapped, we should run through some testing with applications that rely on GraphQL queries to see if there are any bugs that crop up. If all looks good, we can follow the same approach to swap production as well! 🚀

### DIFF
--- a/dosomething-dev/main.tf
+++ b/dosomething-dev/main.tf
@@ -53,7 +53,7 @@ module "graphql_lambda" {
 
   environment = "development"
   name        = "dosomething-graphql-dev"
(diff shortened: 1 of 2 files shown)


DFurnes:
I've applied e3d0e4c successfully. Since it'll still take a few for the certificates to provision, I'm going to plan on making the DNS switch tomorrow morning. 🌞

PR #151, opened 2019-02-12
Add base domain for Footlocker Internal.

### LINKED ISSUE #138: Configure DNS for Foot Locker internal scholarship app.
For previous scholarship apps, we've run these using the client's registrar & DNS (at their request). This meant we had to use an A Record routed to our HAProxy instance (with a static IP).

For the upcoming internal scholarship, the Foot Locker team let us know that we'd been running these on a "free tier" DNS provider, which may not be able to handle a lot of traffic:

> I have access to the registrar level and can update which DNS a domain is pointed at. However, this one currently is pointed at MarkMonitor’s free DNS offering. I would not recommend making use of this as there are hard limits on DNS lookups it will support and no SLA because it is a free service. We should determine which name servers are going to handle the domain. I can then point the domain to those NS(s) and then we’ll likely need the networking team to update the A record on the root and possibly the www. record too. I’m assuming that’s ultimately what you’re trying to do?

Since it was an option this time, we decided to put this on our DNSMadeEasy account.

References [this Slack conversation](https://dosomething.slack.com/archives/C03T8SDMW/p1549468339005000).

DFurnes:
This pull request attaches the base `flscholarship.com` to the Heroku app, so that Heroku knows how to respond to requests that come in sans the `www.` subdomain. (If we configure more applications like this, we'll probably want to add multiple domain support to the `heroku_app` module).

Closes #138.

### DIFF
--- a/longshot/main.tf
+++ b/longshot/main.tf
@@ -50,7 +50,7 @@ module "longshot-footlocker-internal" {
 
 # Attach base domain for FLScholarship.com as well.
 # TODO: Is there a better way to handle ANAMEs?
-resource "heroku_domain" "domain" {
+resource "heroku_domain" "flscholarship_base_domain" {
   app      = "${module.longshot-footlocker-internal.name}"
   hostname = "flscholarship.com"
 }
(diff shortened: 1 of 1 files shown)


DFurnes:
Applied, and confirmed this base domain now loads as expected!

screen shot 2019-02-12 at 3 55 58 pm

PR #152, opened 2019-02-12
Prepare to swap GraphQL to the Lambda backend in production.

DFurnes:
This pull request follows in #149's footsteps and attaches `graphql.dosomething.org` to our production Lambda so we can provision the ACM certificate & CloudFront distribution in advance of swapping real traffic over tomorrow. :v:


DFurnes:
Applied! 🎯

PR #154, opened 2019-02-15
Enable caching on all envs except Preview.

weerd:
This PR enables caching on all Phoenix environments except Phoenix Preview (which is meant to showcase content changes as they happen), so that these envs (Dev and QA) mimic how Production behaves.


DFurnes:
Looks good to me!

```
~ module.dosomething-dev.module.phoenix.module.app.heroku_app.app
    config_vars.0.CONTENTFUL_CACHE:   "false" => "true"

~ module.dosomething.module.phoenix_preview.module.app.heroku_app.app
    config_vars.0.CONTENTFUL_CACHE:   "true" => "false"
```

Applied 7ccd1f7 successfully & merging!

PR #155, opened 2019-02-15
Use 3 queue workers on Northstar production

katiecrane:
We were already using 0 queue workers on development and 1 on QA, this just increases the production queue workers from 1 to 3:

```
~ module.dosomething.module.northstar.module.app.heroku_formation.queue
    quantity:   "1" => "3"
```

Was also getting a bunch of fastly changes when `make plan` ran but @DFurnes suspects that could be due to using the read-only credentials.

### DIFF
--- a/applications/northstar/main.tf
+++ b/applications/northstar/main.tf
@@ -117,7 +117,7 @@ module "app" {
   ignore_web = "${var.environment == "production"}"
 
   # We don't run a queue process on development right now. @TODO: Should we?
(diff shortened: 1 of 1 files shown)


DFurnes:
Yup, confirming I just get the one `heroku_formation.queue` change when I plan on my end. I think we can probably clear up those Fastly diffs by explicitly setting `activate = true` on those resources… I'll make a note to look into that later.

DFurnes:
Applied! Merging this in. 🛠

PR #157, opened 2019-02-19
Remove Heroku GraphQL backend.

DFurnes:
Now that we've had our GraphQL service running on Lambda for a while without issues, I think it's safe to tear down our old Heroku infrastructure for this app. (If we ever need to spin it back up again, it should be as easy as reverting this PR!)

```
Plan: 0 to add, 3 to change, 22 to destroy.
```


DFurnes:
Applied! 🔥

DFurnes:
Ah, some stragglers! 👁 I removed some references to the Heroku pipeline that I'd forgot above.

DFurnes:
Finally, I've renamed the `graphql_lambda` modules to plain ol' `graphql` in 0eae5d2.

PR #158, opened 2019-02-21
Set X-Fastly-Region-Code header on both request and response.

DFurnes:
Two updates to this functionality:

💬 I'd assumed we could read this from the HTML response headers in JavaScript, but it [turns out that's not the case](https://github.com/DoSomething/phoenix-next/pull/1266). This pull request updates our front-end Fastly property to set the `X-Fastly-Region-Code` header on _both_ incoming requests & outgoing responses.

📃 I'd forgotten that we could just set this using the `header` block on the [fastly_service_v1](https://www.terraform.io/docs/providers/fastly/r/service_v1.html) resource, instead of having to write up a custom VCL snippet. Way simpler!

### DIFF
--- a/dosomething-dev/fastly-frontend/main.tf
+++ b/dosomething-dev/fastly-frontend/main.tf
@@ -90,7 +90,7 @@ resource "fastly_service_v1" "frontend-dev" {
     name        = "Region Code"
     type        = "request"
     action      = "set"
-    source      = "geoip.region"
+    source      = "client.geo.region"
     destination = "http.X-Fastly-Region-Code"
   }
 
@@ -98,7 +98,7 @@ resource "fastly_service_v1" "frontend-dev" {
     name        = "Region Code (Debug)"
     type        = "response"
     action      = "set"
-    source      = "geoip.region"
+    source      = "client.geo.region"
     destination = "http.X-Fastly-Region-Code"
   }
 
--- a/dosomething-dev/fastly-frontend/main.tf
+++ b/dosomething-dev/fastly-frontend/main.tf
@@ -86,6 +86,22 @@ resource "fastly_service_v1" "frontend-dev" {
     destination = "http.X-Fastly-Country-Code"
   }
 
+  header {
+    name        = "Region Code"
+    type        = "request"
+    action      = "set"
+    source      = "geoip.region"
+    destination = "http.X-Fastly-Region-Code"
+  }
+
+  header {
+    name        = "Region Code (Debug)"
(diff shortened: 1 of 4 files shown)


mendelB:
@DFurnes Why do we need this property set on incoming requests?

DFurnes:
> @DFurnes Why do we need this property set on incoming requests?

Since we can't access response headers from the JavaScript app, we have to [read them in Laravel and set them on a `window` object](https://github.com/DoSomething/phoenix-next/pull/1266/files#diff-80bda0ca3af455d6354e095203e0199aR44). The way we'd set this up before only set the `X-Fastly-Region-Code` header on the response (after Laravel renders & returns the HTML) so it would read as `null`.

DFurnes:
Huh, this is acting funny. I still only see the `X-Fastly-Country-Code` header on the response when debugging. I checked the [generated VCL](https://manage.fastly.com/configure/services/4W2LIOsS5Yw8DzvZp9iT0z/versions/38/vcl) and [copied it into a Fiddle](https://fiddle.fastlydemo.net/fiddle/643a2538), though and it appends the `X-Fastly-Region-Code` header as expected!

It's also interesting that we use `geoip.country_code` for the country code header (and it works!) since that's not documented anywhere that I can find & actually shows as an "error" in the Fiddle.

DFurnes:
Huh, just broken on www-dev.dosomething.org, though. Works as expected elsewhere...

```
curl -Is https://www-dev.dosomething.org/us | grep x-fastly
x-fastly-country-code: US
```
```
curl -Is https://qa.dosomething.org/us | grep x-fastly
x-fastly-country-code: US
x-fastly-region-code: NY
```

DFurnes:
And we're definitely routing  [www-dev.dosomething.org](https://www-dev.dosomething.org) to Fastly...

```
nslookup www-dev.dosomething.org
Server:		8.8.8.8
Address:	8.8.8.8#53

Non-authoritative answer:
www-dev.dosomething.org	canonical name = osff.map.fastly.net.
Name:	osff.map.fastly.net
Address: 151.101.2.217
```

DFurnes:
Oh, duh. The dev property was set up to handle [staging.dosomething.org](https://staging.dosomething.org)… `www-dev.dosomething.org` was running through the backend property for the dev environment. 😖 That's confusing. I'm going to clean up after merging this in.

```
curl -Is https://staging.dosomething.org/us | grep x-fastly
x-fastly-country-code: US
x-fastly-region-code: NY
```

PR #159, opened 2019-02-21
Use 'dev.dosomething.org' for Phoenix development, and 'preview.dosomething.org' for Phoenix preview.

DFurnes:
We'd been running `staging.dosomething.org` through our Fastly frontend service for the development environment, but I think most people (myself included) were using `www-dev.dosomething.org` to test things out. This pull request updates this app to run through our frontend property via `dev.dosomething.org` for consistency.

We had originally used this domain for local Ashes development and instructed developers to set it to loopback to localhost on their `/etc/hosts` files. I'll drop a note in #announce-tech to make sure everyone has removed that entry once this is all in place.

### DIFF
--- a/dosomething-dev/fastly-backend/main.tf
+++ b/dosomething-dev/fastly-backend/main.tf
@@ -22,10 +22,6 @@ resource "fastly_service_v1" "backends-dev" {
   }
 
   domain {
-    name = "${var.phoenix_domain}"
-  }
-
-  domain {
     name = "${var.rogue_domain}"
   }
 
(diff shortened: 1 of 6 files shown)


dosomething-dev/fastly-backend/main.tf
```diff
@@ -21,10 +21,6 @@ resource "fastly_service_v1" "backends-dev" {
     name = "${var.northstar_domain}"
   }
 
-  domain {
-    name = "${var.phoenix_domain}"
```
DFurnes: We'd previously routed `www-dev.dosomething.org` through the backend property, which doesn't make a ton of sense (hence why this is being removed). I'll set up redirects for `www-dev.dosomething.org` and `staging.dosomething.org` to point to `dev.dosomething.org` once this is applied.

DFurnes:
Applied f5ce839 and confirmed requests are hitting `dosomething-phoenix-dev` now.

DFurnes:
Updated the Redirect URI for [this app's OAuth client](https://admin-dev.dosomething.org/clients/phoenix) & confirmed auth works as expected.

DFurnes:
Updated the nginx config in `/etc/nginx/sites-enabled/dosomething` on the Ashes staging EC2 instance so (the few) Ashes paths we still serve continue to work in this environment. 🐦

DFurnes:
Added a redirect from `www-dev.dosomething.org` to `dev.dosomething.org` in 4efb153.

DFurnes:
Finally, added a redirect from from `www-preview.dosomething.org` to `preview.dosomething.org` in ba67b1d and confirmed that it works as expected! I'll also update the URL used for these in Contentful.

PR #164, opened 2019-03-21
Updating log settings for slow queries

blisteringherb:
This adds log settings so that we're capturing slow queries. We may want to change `log_statement` soon after we gather our slow query info so that we're not missing log entries. This will update the following settings:
`log_min_duration_statement`
`log_statement`
`log_duration`

Note: This will remove the `random_page_cost` and `work_mem` settings we have set up in QA since those aren't committed to master yet.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -264,6 +264,24 @@ resource "aws_db_parameter_group" "quasar-qa" {
     name  = "rds.force_ssl"
     value = "1"
   }
+
+  # Only log queries with a duration longer than this to get slow queries.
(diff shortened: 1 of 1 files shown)


blisteringherb:
It's supposed to make them less noisy. I was double checking the settings and ran across this Stack Overflow that suggested turning that setting to `none`. https://stackoverflow.com/questions/8414172/cannot-get-log-min-duration-statement-to-work/8425131#8425131

sheyd:
Sounds great! Apply and mergey merge!

PR #166, opened 2019-03-27
Disable Synchronous Query to Disk 

sheyd:
Per Sean Hull's recommendation, we don't need to have each transaction committed to disk before they report back as successful, and with the tradeoff of potentially losing a small amount of data for a large performance boost. Testing out in QA first.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -271,10 +271,10 @@ resource "aws_db_parameter_group" "quasar-qa" {
     value = "1000"
   }
 
-  # Temporarily only log slow queries.
+  # Testing turning synchronization off to see if improves performance
   parameter {
-    name  = "log_statement"
-    value = "none"
+    name  = "synchronous_commit"
+    value = "off"
   }
 }
 
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -277,11 +277,21 @@ resource "aws_db_parameter_group" "quasar-qa" {
     value = "none"
   }
 
+  parameter {
+    name  = "log_duration"
+    value = "0"
+  }
+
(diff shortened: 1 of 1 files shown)


sheyd:
```
~ module.quasar.aws_db_parameter_group.quasar-prod
    parameter.#:                         "9" => "8"
    parameter.1092053286.apply_method:   "immediate" => ""
    parameter.1092053286.name:           "log_duration" => ""
    parameter.1092053286.value:          "1" => ""

~ module.quasar.aws_db_parameter_group.quasar-qa
    parameter.#:                         "10" => "9"
    parameter.1092053286.apply_method:   "immediate" => ""
    parameter.1092053286.name:           "log_duration" => ""
    parameter.1092053286.value:          "1" => ""
    parameter.2771588648.apply_method:   "" => "immediate"
    parameter.2771588648.name:           "" => "synchronous_commit"
    parameter.2771588648.value:          "" => "off"
    parameter.3022839578.apply_method:   "immediate" => ""
    parameter.3022839578.name:           "random_page_cost" => ""
    parameter.3022839578.value:          "1" => ""

Plan: 0 to add, 2 to change, 0 to destroy.
```

sheyd:
Despite running `make apply`, it looks like the `log_duration` settings isn't taking for Quasar Prod or QA parameter groups. If I explicitly set them on the RDS console and match the settings in Terraform, the diffs go away. I think this has to do with `log_duration` requiring superuser privileges to be set according to https://www.postgresql.org/docs/10/runtime-config-logging.html#GUC-LOG-MIN-DURATION-STATEMENT.

sheyd:
Confirmed once those settings were explicitly set, state is back to normal.

PR #168, opened 2019-04-03
Use named '/' and '/graphql' routes instead of catch-all proxy.

DFurnes:
This pull request swaps GraphQL (development & QA) to use the `api_gateway` module, rather than the "catch-all" `api_gateway_proxy` that I'd started with (where _every_ route will go to the same Lambda function).

This more clearly lines up to real-world usage (where requests all go to either `/` or `/graphql`) , and opens us up to expose other functions as part of this application (say, in DoSomething/graphql#96).

~Note that I'm temporarily using a `graphql-beta` module so we can roll this change out _only_ for our development & QA environments at first.~ Done! I've tested these changes & updated the standard `applications/graphql` module to apply the same configuration to production.


DFurnes:
Welp! Tested this out (on dev & QA) and didn't think to move the `aws_api_gateway_domain_name` to the new "path" in state, and so it got destroyed & re-created. This ended in a little kerfuffle – AWS won't let you create a new CloudFront distribution when the DNS record still points to the old one.

I'm rolling back the apply (and pending the re-provisioned CloudFront distributions, will try this again)!

DFurnes:
Rolled back & fixed a little mistake in 9bc876a. This time around, remembered to move the domains to their new "path" in state so they don't rebuild, and applied:

```sh
# Ensure we don't rebuild domains & certificates:
terraform state mv module.dosomething-qa.module.graphql.module.gateway.module.api_gateway.aws_api_gateway_domain_name.domain module.dosomething-qa.module.graphql.module.gateway.aws_api_gateway_domain_name.domain
terraform state mv module.dosomething-dev.module.graphql.module.gateway.module.api_gateway.aws_api_gateway_domain_name.domain module.dosomething-dev.module.graphql.module.gateway.aws_api_gateway_domain_name.domain

# Do the things!
make apply
```

Looking good this time! 😌

DFurnes:
I've just moved the domain resource in production so it won't rebuild when we apply this:

```
terraform state mv module.dosomething.module.graphql.module.gateway.module.api_gateway.aws_api_gateway_domain_name.domain module.dosomething.module.graphql.module.gateway.aws_api_gateway_domain_name.domain
```

DFurnes:
Thanks @mendelB! I've applied this change & confirmed that https://graphql.dosomething.org and https://graphql.dosomething.org/graphql are both loading as expected! ⛵️

PR #169, opened 2019-04-04
Add Contentful webhook function.

DFurnes:
This pull request adds another function to our GraphQL app, running the `webhook.handler` handler.

References DoSomething/graphql#96.

### DIFF
--- a/applications/graphql/main.tf
+++ b/applications/graphql/main.tf
@@ -97,6 +97,10 @@ module "app" {
   }
 }
 
+resource "random_string" "webhook_secret" {
+  length = 32
+}
+
 module "contentful_webhook" {
   source = "../../components/lambda_function"
 
@@ -110,6 +114,7 @@ module "contentful_webhook" {
     QUERY_ENV      = "${local.env}"
     CACHE_DRIVER   = "dynamodb"
     DYNAMODB_TABLE = "${module.cache.name}"
+    WEBHOOK_SECRET = "${random_string.webhook_secret.result}"
   }
 }
 
--- a/applications/graphql/main.tf
+++ b/applications/graphql/main.tf
@@ -124,8 +124,11 @@ module "gateway" {
   name   = "${var.name}"
   domain = "${var.domain}"
 
-  functions     = ["${module.app.arn}", "${module.contentful_webhook.arn}"]
-  root_function = "${module.app.invoke_arn}"
+  functions_count = 2
+  functions       = ["${module.app.arn}", "${module.contentful_webhook.arn}"]
+  root_function   = "${module.app.invoke_arn}"
+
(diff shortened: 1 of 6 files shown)


applications/graphql/main.tf
```diff
@@ -97,20 +97,41 @@ module "app" {
...
+}
+
 module "gateway" {
   source = "../../components/api_gateway"
 
   name   = "${var.name}"
   domain = "${var.domain}"
 
-  functions     = ["${module.app.arn}"]
+  functions     = ["${module.app.arn}", "${module.contentful_webhook.arn}"]
```
weerd: what's an `arn`? 🤔

DFurnes: Amazon's [silly name for IDs](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html). 🤷‍♂️

weerd: ah! 🤪

DFurnes:
Alas, this became a wee bit more complicated –

1. Added permissions so that both Lambda functions can access the DynamoDB table in b3af523. This was a relatively straightforward change to make the `role` argument into a `roles` array in the `dynamodb_cache` component (which is only used here).
2. Fixed a Terraform 0.11 issue with dynamic lists used for the [`count` argument](https://www.terraform.io/docs/configuration/resources.html#count-multiple-resource-instances) by passing a (manually maintained, but "validated") count integer argument alongside it. This hack can be removed once we're on Terraform 0.12.

DFurnes:
Confirmed that this PR now makes a ton of new resources for the new Lambdas & their permissions/routes, but does **not** delete or change any existing resources.

```
Plan: 74 to add, 0 to change, 0 to destroy.
```

DFurnes:
I ran into an interesting snag yesterday – by default, the API Gateway deployment [doesn't update](https://github.com/hashicorp/terraform/issues/6613) when routes change (and so those routes are not accessible by real traffic, even though they show up in the AWS console).

It seems like this can be fixed by setting a `variable` or `stage_description` on the deployment, which forces a rebuild. I tried that out yesterday and ran into conflicts with not being able to delete an active stage (reasonable), since Terraform always tries to delete an existing resource before building its replacement.

This (hopefully) can be fixed in d6e3d2e with the [`create_before_destroy`](https://www.terraform.io/docs/configuration/resources.html#lifecycle-lifecycle-customizations) lifecycle argument, which I've added in d6e3d2e.

DFurnes:
Alright, confirmed that this is working (with the default "hello world" function), and the new deployment is automatically created without issues across the board. 🕺 

```
curl -X POST https://graphql-dev.dosomething.org/contentful
<p>Hello world!</p>
```

PR #171, opened 2019-04-09
Add basic Chompy application template & generate API keys.

DFurnes:
This pull request adds a super bare-bones Chompy application template & sets the `IMPORTER_API_KEY` environment variable using a [`random_string`](https://www.terraform.io/docs/providers/random/r/string.html) resource. This allows us to safely generate a secure key. If we wanted to re-generate this later (say, if it leaked), we'd just run:

```
terraform taint -module=dosomething-qa.chompy random_string.importer_api_key
```

We were a little pressed for time, so we only set up the apps themselves (without bothering to hook up the domain, logging, database, S3 bucket, queue, yadda yadda). This is something we should totally revisit when we have time during a cleanup sprint later in the quarter!

### DIFF
--- a/applications/chompy/main.tf
+++ b/applications/chompy/main.tf
@@ -10,6 +10,10 @@ variable "name" {
   description = "The application name."
 }
 
+resource "random_string" "importer_api_key" {
+  length = 32
+}
+
 # TODO: Use our 'heroku_app' module to configure more of this!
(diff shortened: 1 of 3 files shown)


DFurnes:
We've applied these changes & verified they look good while pairing on this work! ⛵️

PR #173, opened 2019-04-25
Add 'Vary' header to S3 responses that are missing it.

DFurnes:
This pull request should fix a [rather frustrating oversight](https://forums.aws.amazon.com/message.jspa?messageID=804087) in S3's CORS support, where the `Vary` header is omitted for requests that don't have a CORS header (and therefore Fastly, CloudFront, or whoever will then just cache the CORS-less response for everyone).

Here's the plan for this change:

```
~ module.dosomething.module.fastly-frontend.fastly_service_v1.frontend
    condition.#:                            "3" => "4"
    condition.3035584417.name:              "" => "response-assets"
    condition.3035584417.priority:          "" => "10"
    condition.3035584417.statement:         "" => "req.http.host == "assets.dosomething.org""
    condition.3035584417.type:              "" => "RESPONSE"
    header.#:                               "4" => "5"
    header.3732783554.action:               "" => "set"
    header.3732783554.destination:          "" => "http.Vary"
    header.3732783554.ignore_if_set:        "" => "true"
    header.3732783554.name:                 "" => "S3 Vary"
    header.3732783554.priority:             "" => "100"
    header.3732783554.regex:                "" => "<computed>"
    header.3732783554.response_condition:   "" => "response-assets"
    header.3732783554.source:               "" => "Origin, Access-Control-Request-Headers, Access-Control-Request-Method"
    header.3732783554.substitution:         "" => "<computed>"
    header.3732783554.type:                 "" => "response"

Plan: 0 to add, 1 to change, 0 to destroy.
```

References [#165498230](https://www.pivotaltracker.com/story/show/165498230).

### DIFF
--- a/dosomething/fastly-frontend/main.tf
+++ b/dosomething/fastly-frontend/main.tf
@@ -111,7 +111,6 @@ resource "fastly_service_v1" "frontend" {
     action             = "set"
     destination        = "http.Vary"
     source             = "\"Origin, Access-Control-Request-Headers, Access-Control-Request-Method\""
-    ignore_if_set      = true
   }
 
   header {
--- a/dosomething/fastly-frontend/main.tf
+++ b/dosomething/fastly-frontend/main.tf
@@ -24,8 +24,8 @@ resource "fastly_service_v1" "frontend" {
   }
 
   condition {
-    type      = "RESPONSE"
-    name      = "response-assets"
+    type      = "CACHE"
+    name      = "cache-assets"
     statement = "req.http.host == \"${var.assets_domain}\""
   }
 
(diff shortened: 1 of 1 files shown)


sheyd:
What an odd oversight :/

DFurnes:
Had to make two small tweaks: adding another round of quotes in b2daf6b since the `source` argument is actually a VCL snippet & removing my overly cautious `ignore_if_set` in b2daf6b since sometimes it seems like Amazon decides to just send a willy-nilly `vary: Accept-Encoding`:

```
λ curl -sI "https://assets.dosomething.org/us/fonts/covered-grace/CoveredGrace-Regular.ttf" | grep "vary"
vary: Accept-Encoding

DFurnes:
Oops, and finally swapping this to be a `cache` condition and header rule, since [that'll affect the response _before_ Fastly caches it rather than after](https://docs.fastly.com/guides/conditions/about-conditions#types-of-conditions-and-when-you-can-use-them). Everything seems to be working now!

PR #174, opened 2019-04-29
Add a 'force pass' rule for authenticated requests.

DFurnes:
This pull request adds a "force pass" rule that tells Fastly to send along authenticated requests to the backend (so we can cache page markup in Fastly for anonymous folks, but not serve that same response to folks who are logged-in).

References DoSomething/phoenix-next#1395 and [#165515770](https://www.pivotaltracker.com/story/show/165515770).

### DIFF
--- a/dosomething-dev/fastly-frontend/main.tf
+++ b/dosomething-dev/fastly-frontend/main.tf
@@ -44,9 +44,12 @@ resource "fastly_service_v1" "frontend-dev" {
   }
 
   condition {
-    type      = "REQUEST"
-    name      = "is-authenticated"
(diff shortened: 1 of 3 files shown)


DFurnes:
Applied this to the development environment, and will promote to QA/production pending testing!

DFurnes:
Confirmed that requests with `laravel_session=` are (still) uncached, and static assets with `laravel_session=` continue to be cached with this change. Once DoSomething/phoenix-next#1395 is approved and merged, we should start to see anonymous page-views be served from cache.

Since everything seems to be working well, I'm going to promote this to QA and then production.

DFurnes:
Applied to QA & production! Going to merge this in. 🎯

PR #175, opened 2019-04-30
Adding new PG11 parameter group to prepare for upgrade

blisteringherb:
This PR adds two new Postgres 11 parameter groups to prepare for our upgrade. We'll need to run a `make apply` to add these before running the upgrade on either QA or Prod, but there shouldn't be any impact to creating both of these together.

After this PR is approved I'll run this, and then put in a separate PR to upgrade QA to Postgres 11.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -294,9 +294,147 @@ resource "aws_db_parameter_group" "quasar-qa" {
   #  }
 }
 
+resource "aws_db_parameter_group" "quasar-qa-pg11" {
+  name   = "quasar-qa-pg11"
+  family = "postgres11"
+
+  # Required for DMS CDC Replication:
+  # https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.v10
+  parameter {
+    name         = "rds.logical_replication"
+    value        = "1"
+    apply_method = "pending-reboot"
+  }
+
+  # Required for DMS CDC Replication:
+  # https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.v10
+  parameter {
+    name         = "max_logical_replication_workers"
+    value        = "500"
+    apply_method = "pending-reboot"
+  }
+
+  # Recommended by PGTuner tool: https://pgtune.leopard.in.ua/#/
+  # Sets effective RAM available to PG Query planner before using disk.
+  parameter {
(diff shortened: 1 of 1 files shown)


quasar/main.tf
```diff
@@ -294,6 +294,78 @@ resource "aws_db_parameter_group" "quasar-qa" {
   #  }
 }
 
+resource "aws_db_parameter_group" "quasar-qa-pg11" {
+  name   = "quasar-qa-pg11"
+  family = "postgres11"
+
+  # Required for DMS CDC Replication:
+  # https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.v10
```
rapala61: Just wondering this comment should point to [CHAP_Source.PostgreSQL.RDSPostgreSQL](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.RDSPostgreSQL) instead. This section holds the recommendation to use this parameter.

> option

ameter

quasar/main.tf
```diff
@@ -294,6 +294,78 @@ resource "aws_db_parameter_group" "quasar-qa" {
...
+  parameter {
+    name         = "max_logical_replication_workers"
+    value        = "500"
+    apply_method = "pending-reboot"
+  }
+
+  # Recommended by PGTuner tool: https://pgtune.leopard.in.ua/#/
```
rapala61: Just curious if you think we should document the assumptions we input in this tool to get the recommendations.

rapala61:
Just left some comments to feel like I belong hehe. From what I can tell, we really just grabbed the old parameters and pasted them to a new parameter group with a different name and 1 or 2 more parameters for the new PostgreSQL 11 instance right?

quasar/main.tf
sheyd: We can add some more nuance here, but the assumption I think started out as "we don't know what the hell we're doing, this is a good starting point." I think we've noted that in the original setup ticket for this in Pivotal which I think is a better place for extended comments. We could possibly link to that ticket

quasar/main.tf
sheyd: The original link pointed to the Postgres 10 link, so might be worth an update, but we're deprecating DMS soon, so we'll probably remove the code. leaving it up to Morgan.

blisteringherb: That 10.x link points to the documentation for version 10 and later, so we can leave that for the moment.

PR #176, opened 2019-05-02
Updating settings to run QA PG11 update immediately

blisteringherb:
This PR updates the settings for QA Quasar so that we can upgrade immediately to Postgres 11. Without this code I was running into the following error:

```
aws_db_instance.quasar-qa: Error modifying DB Instance quasar-qa: InvalidParameterCombination: The AllowMajorVersionUpgrade flag must be present when upgrading to a new major version.
```

This PR adds the `allow_major_version_upgrade` and `apply_immediately` flags for Quasar QA.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -523,7 +523,7 @@ resource "aws_db_instance" "quasar-qa" {
   name                            = "quasar"
   username                        = "${data.aws_ssm_parameter.qa_username.value}"
   password                        = "${data.aws_ssm_parameter.qa_password.value}"
-  parameter_group_name            = "${aws_db_parameter_group.quasar-qa.id}"
+  parameter_group_name            = "${aws_db_parameter_group.quasar-qa-pg11.id}"
   vpc_security_group_ids          = ["${aws_security_group.rds.id}"]
   deletion_protection             = true
   storage_encrypted               = true
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -519,6 +519,7 @@ resource "aws_db_instance" "quasar-qa" {
   engine                          = "postgres"
   engine_version                  = "11.1"
   instance_class                  = "db.m5.2xlarge"
+  allow_major_version_upgrade     = true
   name                            = "quasar"
   username                        = "${data.aws_ssm_parameter.qa_username.value}"
   password                        = "${data.aws_ssm_parameter.qa_password.value}"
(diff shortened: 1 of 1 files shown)


quasar/main.tf
```diff
@@ -519,6 +519,7 @@ resource "aws_db_instance" "quasar-qa" {
   engine                          = "postgres"
   engine_version                  = "11.1"
   instance_class                  = "db.m5.2xlarge"
+  allow_major_version_upgrade     = true
   name                            = "quasar"
   username                        = "${data.aws_ssm_parameter.qa_username.value}"
   password                        = "${data.aws_ssm_parameter.qa_password.value}"
```
sheyd: I think the `parameter_group_name` here needs to be `"${aws_db_parameter_group.quasar-qa-pg11.id}"`.

quasar/main.tf
```diff
@@ -519,17 +519,19 @@ resource "aws_db_instance" "quasar-qa" {
...
-  parameter_group_name            = "${aws_db_parameter_group.quasar-qa.id}"
+  parameter_group_name            = "${aws_db_parameter_group.quasar-qa-pg11.id}"
   vpc_security_group_ids          = ["${aws_security_group.rds.id}"]
   deletion_protection             = true
   storage_encrypted               = true
   copy_tags_to_snapshot           = true
   monitoring_interval             = "10"
   publicly_accessible             = true
   enabled_cloudwatch_logs_exports = ["postgresql"]
+  apply_immediately               = true
```
DFurnes: Let's not use `apply_immediately` in our Terraform config, since it's nice to be able to safely queue up a change and let it get "committed" during the maintenance window if it requires downtime. If you want to "force" a change, you can use the AWS CLI to apply queued changes immediately after running `make apply`:

```
aws rds modify-db-instance --db-instance-identifier quasar-qa --apply-immediately
```

sheyd: @DFurnes What's the best place to document this (if we haven't already)? I mistakenly told Morgan we should put that config in, then remove it in another PR once the upgrade was complete, but totally makes sense what you're saying. I think we've had this conversation before and I'd like to make sure we can capture this somewhere for future DB updates!

DFurnes: I think this would make sense to document [once we extract a `postgres_warehouse` component](https://www.pivotaltracker.com/story/show/165499203) for these resources (following the example of how [`api_gateway`](https://github.com/DoSomething/infrastructure/tree/master/components/api_gateway) and [`lambda_function`](https://github.com/DoSomething/infrastructure/tree/master/components/lambda_function) are documented.)

PR #177, opened 2019-05-08
Swap backend to Terraform Enterprise.

DFurnes:
This pull request swaps our backend from S3/DynamoDB to [Terraform Enterprise's free tier](https://www.terraform.io/docs/enterprise/free/overview.html). This allows us to easily see when changes were made (rather than having to dig through the versioned S3 file) & will open us up to cool new features like remote plans & applies in the future.

**Note that any engineers who want to run Terraform will need to [create a Terraform.io account](https://app.terraform.io/account/new) and be invited to our team in order to run plans or applies in the future.** 👈

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Next, **make a plan** to find out how it will affect the current state of the sy
 make plan
 ```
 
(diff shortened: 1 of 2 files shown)


sheyd:
I'm SO EXCITED ABOUT THIS!!!!


image

PR #178, opened 2019-05-08
Don't route traffic to Ashes.

DFurnes:
This pull request removes the Ashes backends and routing rules, because well, we're not sending traffic there anymore! I'll follow up with decommissioning the actual infrastructure next sprint. 👋


DFurnes:
> Should we also remove the ashes references in the `main.tf` files? https://git.io/fjcNX

Yes! I've removed the (Fastly) references from the dev & QA environments, and after testing that that works as expected I'll do the same for production. We can remove each environment's `ashes` module (with the EC2 instances) when we remove the corresponding infrastructure next sprint.

mshmsh5000:
Sounds super great! This is exciting 🎉

DFurnes:
I just applied 97a799e ([diff](https://app.terraform.io/app/dosomething/workspaces/infrastructure/states/sv-2s1AKUDMhoBqwgJn)) and confirmed everything works as expected. There were a few (unrelated) errors for Quasar that it seems will be addressed once #176 is merged.

PR #179, opened 2019-05-08
Don't route traffic to Ashes. (Production)

DFurnes:
This pull request promotes the changes from #178 to production – removing the Ashes backend from Fastly, and the condition & custom rules that sent traffic there. Once we're confident that everything worked as expected with this change on Development/QA, we can apply this!


DFurnes:
Applied! ([diff](https://app.terraform.io/app/dosomething/workspaces/infrastructure/states/sv-Ne7DyNi9AxkQCQv2))

PR #180, opened 2019-05-10
Temporary: log user agents on incoming traffic.

DFurnes:
I'm curious if we can learn a bit more about these traffic spikes in DoSomething/internal#478 by temporarily logging the `User-Agent` header on incoming requests following our next broadcast.


DFurnes:
Applied 0052eba and confirmed we're getting user-agents in the logs now!

PR #182, opened 2019-05-15
Updating Postgres version to reflect reality

blisteringherb:
It seems that trying to run a major version upgrade through Terraform has no result, so I had to update through the GUI and then make the changes in Terraform to reflect which version we're on. This PR just updates the version from 11.1 to 11.2 which is the most recent version available.


DFurnes:
Yeah, this should be something you can "force apply" via the CLI.

Going to merge this in so I don't revert your changes when I apply #183!

PR #183, opened 2019-05-15
Clean up log formatting.

DFurnes:
This pull request ~removes the temporary `user-agent` field added in #181, and~ swaps elapsed time to be logged in milliseconds for readability & consistency with other systems, like [Heroku's router logs](https://devcenter.heroku.com/articles/http-routing#heroku-router-log-format).

### DIFF
--- a/dosomething-dev/fastly-backend/main.tf
+++ b/dosomething-dev/fastly-backend/main.tf
@@ -161,18 +161,9 @@ resource "fastly_service_v1" "backends-dev" {
   }
 
   papertrail {
-    name               = "northstar-dev"
-    address            = "${element(split(":", var.papertrail_destination), 0)}"
-    port               = "${element(split(":", var.papertrail_destination), 1)}"
-    format             = "${var.papertrail_log_format}"
-    response_condition = "response-northstar-dev"
-  }
-
-  papertrail {
-    name               = "rogue-dev"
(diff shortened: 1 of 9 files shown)


DFurnes:
After testing this works as expected, I'll update our other Fastly log formats to use `ms=` as well.

mshmsh5000:
Looks good. I can't imagine we'd never want to query across Heroku and Fastly logs, otherwise it might make more sense to standardize this around `service=[n]ms`. I assume this is just for easier readability by humans.

DFurnes:
Yeah, but I do kinda like using `service=[n]ms` just for consistency... hmmm. 🤔

DFurnes:
Rebased against `master` so I don't overwrite changes from #182.

DFurnes:
I rebased against `master` & rolled out this new log format on dev & QA (and DRY'd things up by defining in each of the "root" environments). I'm going to keep `user-agent` for now since that's still been useful for debugging, and doesn't take up too much log volume.

DFurnes:
I added a `app=` field to log which backend handles a request (e.g. `app=F_dosomething_phoenix_qa`) and renamed the `program` field to just `frontend` or `backend` so we don't need multiple log connections per Fastly property.

I'll update any Papertrail saved searches that rely on the previous structure.

DFurnes:
Applied! Going to merge this in.

PR #184, opened 2019-05-16
Stop managing Rogue's web dynos with Terraform.

DFurnes:
This pull request sets the `ignore_web` flag for Rogue's production application, because we'll be handling autoscaling for this app with [HireFire](https://hirefire.io) (like we do for Northstar). (I've also removed the `web_scale` override since `1` is the default.)


DFurnes:
I persisted [this manual change to Quasar QA](https://dosomething.slack.com/archives/C03T8SDJJ/p1557942305047200) in a9bfcae so applying this won't roll back that change.

PR #186, opened 2019-05-22
A little more log formatting cleanup.

DFurnes:
This pull request adds quotes around the `cache` log field, so shielded requests (e.g. `cache="HIT, MISS"` are a little easier to read and search for. I've also renamed the `X-Origin-Name` header to `X-Application-Name` for clarity (since it's really about which app/backend serves the request).

Finally, I've reformatted that header to be more consistent with the "real" application names (for example, changing `F_dosomething_phoenix_qa` into `dosomething-phoenix-qa`). ([fiddle demo](https://fiddle.fastlydemo.net/fiddle/0d90256a))


DFurnes:
Applied!

PR #187, opened 2019-05-22
Fix issue where app name wasn't completely transformed.

DFurnes:
Whoops. The [`regsub`](https://docs.fastly.com/vcl/functions/regsub/) helper only makes a single replacement, so I need to use [`regsuball`](https://docs.fastly.com/vcl/functions/regsuball/) here! (This fixes some awkward logs, like `app=dosomething-northstar_qa`.... so close!!)


DFurnes:
Applied & confirmed this is getting transformed correctly now!

PR #188, opened 2019-05-31
updates chompy/main.tf file with SoftEdge and CallPower keys

chloealee:
updates chompy/main.tf file with SoftEdge and CallPower keys with new changes in Chompy [here](https://github.com/DoSomething/chompy/pull/93)

### DIFF
--- a/applications/chompy/main.tf
+++ b/applications/chompy/main.tf
@@ -10,7 +10,11 @@ variable "name" {
   description = "The application name."
 }
 
-resource "random_string" "importer_api_key" {
+resource "random_string" "callpower_api_key" {
+  length = 32
+}
+
+resource "random_string" "softedge_api_key" {
   length = 32
 }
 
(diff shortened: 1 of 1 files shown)


applications/chompy/main.tf
```diff
@@ -24,7 +24,8 @@ resource "heroku_app" "app" {
   }
 
   config_vars = {
-    IMPORTER_API_KEY = "${random_string.importer_api_key.result}"
+    CALLPOWER_API_KEY = "${random_string.importer_api_key.result}"
+    SOFTEDGE_API_KEY = "${random_string.importer_api_key.result}"
```
DFurnes: We'll want these to be different, so let's rename `importer_api_key` to `random_string.callpower_api_key` above & then create a new `random_string.softedge_api_key` to generate the secret for `SOFTEDGE_API_KEY`.

chloealee: ah sorry about that @DFurnes ! I thought that would be called twice and generate two different strings each time. Updated!

DFurnes:
Added 03957b0 so that this won't break production (until we deploy DoSomething/chompy#93), moved the existing `importer_api_key` resources to the new `callpower_api_key` path so they won't regenerate, and applied!

chloealee:
Thanks @DFurnes !!

PR #190, opened 2019-06-17
Update provider versions.

DFurnes:
This pull request updates our Terraform provider versions to the latest available release, paving the way for [#166288730](https://www.pivotaltracker.com/story/show/166288730) in an upcoming sprint. Notably:

- The newer AWS provider adds support for Performance Insights on RDS instances, so I've gone ahead and set that to `true` for Quasar so it isn't reverted to the default (disabled) state.
 - The newer AWS provider changes the `distribution` param for [cloudwatch_log_subscription_filter](https://www.terraform.io/docs/providers/aws/r/cloudwatch_log_subscription_filter.html) to be set to `""` (as expected, since these aren't targeting Amazon Kinesis).
- The newer Heroku provider updates [heroku_app](https://www.terraform.io/docs/providers/heroku/r/app.html)'s `config_vars` to be stored as a map, rather than a list containing a map (which was always awkward). This results in a pretty hefty diff for "setting" the new values, but should be no effective change.
- The new Heroku provider also introduces a `sensitive_config_vars` parameter for [heroku_app](https://www.terraform.io/docs/providers/heroku/r/app.html) that we can move passwords & keys into so they're hidden in diffs.

### DIFF
--- a/components/heroku_app/main.tf
+++ b/components/heroku_app/main.tf
@@ -147,7 +147,7 @@ resource "heroku_app" "app" {
     name = "dosomething"
   }
 
-  config_vars = ["${merge(local.config_vars[var.framework], var.config_vars)}"]
+  config_vars = "${merge(local.config_vars[var.framework], var.config_vars)}"
 
   buildpacks = "${local.buildpacks[var.framework]}"
 
--- a/main.tf
+++ b/main.tf
@@ -37,7 +37,7 @@ terraform {
 # in front of (most of) our services & handles caching,
 # backend-routing, geolocation, redirects, etc.
 provider "fastly" {
-  version = "~> 0.4"
+  version = "~> 0.6"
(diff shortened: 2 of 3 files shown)


DFurnes:
It's worth noting that we're still waiting on an updated MySQL provider ([pending release](https://github.com/terraform-providers/terraform-provider-mysql/blob/master/CHANGELOG.md)) & Fastly provider (still under development) before we can update our codebase to 0.12.

mshmsh5000:
Do we want to ask any particular timeline questions of our Fastly contacts?

DFurnes:
@mshmsh5000 It looks like there is [an active pull request](https://git.io/fjaXb) to add support, but it couldn't hurt to mention it to our contact there so they know folks are interested! 😸

DFurnes:
I've applied this pull request & confirmed nothing broke. 👼

PR #191, opened 2019-06-25
Upgrade MySQL provider.

DFurnes:
My mind is getting fried on Cypress, so I took a brief detour to check in on providers for [#166288730](https://www.pivotaltracker.com/story/show/166288730). After this, we've just got one more ([Fastly](https://git.io/fjr2N)) to go! 🤞 I've also set the `distribution` argument on our `aws_cloudwatch_log_subscription_filter` resources, since otherwise they're always marked as "dirty" in plans/applies.


DFurnes:
Ooops, thanks @mshmsh5000! I forgot to push up 654b451 with that change. 🙈

mshmsh5000:
👍

PR #195, opened 2019-07-12
Upgrade to Terraform 0.12.

DFurnes:
This pull request updates our infrastructure to Terraform 0.12! ✨

I'd recommend reviewing this commit-by-commit:

🗺 I swapped "implicit" relative module paths to explicit ones, e.g. `fastly-frontend` to `./fastly-frontend`, as required by Terraform 0.12. d47833e

💲 I escaped Fastly's log format string, which otherwise conflicted with Terraform's new [string templating directives](https://www.terraform.io/docs/configuration/expressions.html#directives). From what I understand, Fastly should end up with the same format string when this is applied. f1d5f58

🆙 Ran the included `terraform 0.12upgrade` command to update to the new HCL 2 syntax. For the most part, this consisted of just unquoting previously required `"${…}"` around variables. b82aca0

📦 I extracted the "backup" replication bucket from `s3_bucket`, and instead pass a target `s3_bucket` resource as an argument to the module. See full explanation below. 6f0fc24

⛰ Removes [Terraform Landscape](https://github.com/coinbase/terraform-landscape) from our `make plan` command, since it [doesn't support Terraform 0.12](https://git.io/fjX9P) and arguably isn't necessary anymore. cf461aa

🔒 I had locked our Heroku provider version in a966090 due to a [bug](https://git.io/fjXe8), but was able to remove that override in 3a45bc3 since the issue [was fixed](https://github.com/terraform-providers/terraform-provider-heroku/releases/tag/v2.0.3).


DFurnes:
Documenting for future reference: When planning out 438b233, I ran into a [Terraform 0.12 bug](https://git.io/fjX9T) with moving an existing resource in state into a new module (in this case, moving the existing `rogue.storage.aws_s3_bucket.backup` into the extracted `rogue_backup.aws_s3_bucket.backup`.

I was able to work around the error by following [this suggestion](https://git.io/fjX9m) to create a `null_resource` in the new module & then run an "apply" for that single resource (so we get the new module into our state without any of the other "pending" changes in this PR):

```sh
echo 'resource "null_resource" "null" {}' > components/s3_backup_bucket/main.tf
terraform apply -target=module.dosomething.module.rogue_backup.null_resource.null
```

After using the null resource to create the module in state, `terraform state mv` works as expected.

DFurnes:
Here's some more context for 438b233 since it's the "meatiest" commit in this upgrade:

We create a replicated bucket for Rogue's files as a way to avoid losing data if there were a catastrophic failure in Amazon's US East region, where our production resources live. Frustratingly, in order to create resources in another region, you need to make a [second `aws` provider](https://github.com/DoSomething/infrastructure/blob/2f3eb34be9cd42b606f9d7325ba7e7b8843319e2/main.tf#L65-L71), which we alias as `aws.west`.

Since Terraform 0.12 [no longer implicitly passes aliased providers](https://www.terraform.io/docs/configuration/modules.html#implicit-provider-inheritance) to child modules, we'd now need to add an explicit `providers` argument and "declaration" (like [this](https://github.com/DoSomething/infrastructure/blob/2f3eb34be9cd42b606f9d7325ba7e7b8843319e2/main.tf#L109-L112) and [this](https://github.com/DoSomething/infrastructure/blob/2f3eb34be9cd42b606f9d7325ba7e7b8843319e2/dosomething/main.tf#L1-L2)) to every single submodule that could potentially create a resource in the AWS West region… and since most parts of our infrastructure have at least some dependency on `s3_module`, that's almost everything… bleh!!

To avoid that cascading change across the board, I've instead created the backup bucket as a standalone module in our production `dosomething/main.tf` module. This module is then provided the `aws.west` provider, and we pass a reference to the bucket it creates around (to other modules which are none the wiser).

DFurnes:
@mshmsh5000 @sheyd Alright, this behemoth is ready for review! 🐳 I'd strongly recommend reviewing this commit-by-commit, since a lot of it is just noise from the updated syntax.

DFurnes:
I'm going to apply the changes in this PR to our dev & QA environments for testing. 💂

DFurnes:
Hmm, not quite there on the log formatting it seems:

```
Jul 15 16:37:05 fastly-dev frontend [15/Jul/2019:20:37:04 +0000] 'GET /favicon.ico?v1 HTTP/1.1' status=200 app=%{X-Application-Name}o cache="%{X-Cache}o" country=%{X-Fastly-Country-Code}o ip="65.246.187.18" user-agent="%{User-Agent}i" service=%{time.elapsed.msec}Vms
```

Gonna look into how we can make both Terraform & Fastly happy! 🙈

DFurnes:
Oh my god, no wonder! Somehow an extra `%` snuck in to the log format (which also explains why this was showing up as a diff in the first place). I removed it in 838ec45 / bc1f912 and re-applied the Fastly changes against dev, and confirmed that logs are now writing as expected.

DFurnes:
One change I'm particularly excited about is that Terraform now stringifies booleans way better:

```
# module.dosomething-qa.module.northstar.module.app.heroku_app.app will be updated in-place
  ~ resource "heroku_app" "app" {
      ~ config_vars           = {
          ~ "DS_ENABLE_PASSWORD_GRANT"   = "0" -> "false"
          ~ "DS_ENABLE_RATE_LIMITING"    = "1" -> "true"
        }
    }
```

DFurnes:
One last fix to make sure our backup bucket keeps its `private` ACL in 4a64a7f (and a good reminder that we need to prioritize swapping this default option in [#165980718](https://www.pivotaltracker.com/story/show/165980718)).

sheyd:
Forgot to add, AMAZING JOB VIKING @DFurnes !!!!!

Angry Viking

PR #200, opened 2019-09-04
Adding secure password generator and Microsoft AD domain setup for VPN.

sheyd:
This PR accomplishes the following:
* Adds data resource for generated ACM certificate that was manually provisioned from AWS Console. Currently not used for anything, merely here for future building blocks.
* Adds random password generator with [code we already use](https://github.com/DoSomething/infrastructure/blob/master/components/mariadb_instance/main.tf#L95-L100) for the MariaDB master password generator. This is used to set strong admin password for Active Directory service.
* Sets up standard AWS Active Directory service for use with AWS Client VPN.

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -475,3 +475,28 @@ resource "aws_db_instance" "quasar" {
   performance_insights_enabled    = true
   enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
 }
+
+resource "random_string" "admin_password" {
+  # Adding to generate secure password for VPN AD Domain.
+  length = 24
+
+  # We can't use '@' or '$' in MySQL passwords.
+  override_special = "!#%&*()-_=+[]{}<>:?"
+}
+
+
+data "aws_acm_certificate" "vpn-cert" {
+  domain = "vpn.d12g.co"
+}
+
+resource "aws_directory_service_directory" "vpn-ad" {
+  name     = "ad.d12g.co"
+  password = random_string.admin_password.result
+  edition  = "Standard"
+  type     = "MicrosoftAD"
+
+  vpc_settings {
+    vpc_id     = "${aws_vpc.vpc.id}"
+    subnet_ids = ["${aws_subnet.subnet-a.id}", "${aws_subnet.subnet-b.id}"]
+  }
+}
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -478,10 +478,8 @@ resource "aws_db_instance" "quasar" {
 
 resource "random_string" "admin_password" {
   # Adding to generate secure password for VPN AD Domain.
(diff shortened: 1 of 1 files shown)


sheyd:
Output of `make plan`:
```
Terraform will perform the following actions:

  # module.quasar.aws_directory_service_directory.vpn-ad will be created
  + resource "aws_directory_service_directory" "vpn-ad" {
      + access_url        = (known after apply)
      + alias             = (known after apply)
      + dns_ip_addresses  = (known after apply)
      + edition           = "Standard"
      + enable_sso        = false
      + id                = (known after apply)
      + name              = "ad.d12g.co"
      + password          = (sensitive value)
      + security_group_id = (known after apply)
      + short_name        = (known after apply)
      + size              = (known after apply)
      + type              = "MicrosoftAD"

      + vpc_settings {
          + subnet_ids = [
              + "subnet-165c315f",
              + "subnet-be42f093",
            ]
          + vpc_id     = "vpc-4e8dc229"
        }
    }

  # module.quasar.random_string.admin_password will be created
  + resource "random_string" "admin_password" {
      + id               = (known after apply)
      + length           = 24
      + lower            = true
      + min_lower        = 0
      + min_numeric      = 0
      + min_special      = 0
      + min_upper        = 0
      + number           = true
      + override_special = "!#%&*()-_=+[]{}<>:?"
      + result           = (known after apply)
      + special          = true
      + upper            = true
    }

Plan: 2 to add, 0 to change, 0 to destroy.
```

quasar/main.tf
```diff
@@ -475,3 +475,28 @@ resource "aws_db_instance" "quasar" {
   performance_insights_enabled    = true
   enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
 }
+
+resource "random_string" "admin_password" {
+  # Adding to generate secure password for VPN AD Domain.
+  length = 24
+
+  # We can't use '@' or '$' in MySQL passwords.
+  override_special = "!#%&*()-_=+[]{}<>:?"
```
DFurnes: We can remove this override, since this isn't making a MySQL password, right?

sheyd: AD passwords have a funky history of limits and complexity, you know [like this announcement from this year](https://techcommunity.microsoft.com/t5/Azure-Active-Directory-Identity/Removal-of-the-16-character-limit-for-passwords-in-Azure-AD/ba-p/565275) 🙄. I think 24 character length with no special characters makes sense, but can definitely get into the password complexity limits/details if you think it's worth it.

DFurnes: Okay, sure. If the goal is to include no special characters, could you change this to `special = false` and remove the comment referencing MySQL?

sheyd: Can do!

PR #203, opened 2019-09-06
Set ISO-3166 region & zip-code headers via Fastly.

DFurnes:
This pull request updates our Fastly properties to dynamically create request & response header rules based on a dictionary mapping (using [dynamic blocks](https://www.terraform.io/docs/configuration/expressions.html#dynamic-blocks)), and adds the new zip-code header that we need for [#168320231](https://www.pivotaltracker.com/story/show/168320231).

### DIFF
--- a/dosomething-dev/fastly-frontend/main.tf
+++ b/dosomething-dev/fastly-frontend/main.tf
@@ -5,8 +5,10 @@ variable "papertrail_log_format" {}
 
 locals {
   headers = {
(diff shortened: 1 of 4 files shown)


DFurnes:
This has been tested on dev/QA & is ready for review!

DFurnes:
Added a tiny fix in b40e512 to fix an issue reading a data resource [that was deleted this AM](https://dosomething.slack.com/archives/C03T8SDJJ/p1568040196011500).

After that, successfully applied & confirmed everything is working well.

PR #206, opened 2019-09-16
Don't cache 404 responses on Northstar (or other backends).

DFurnes:
This pull request updates our backend Fastly services (including Northstar) to _not_ cache 404 responses, since we're accidentally keeping more transient Heroku errors along longer than they should be.

### DIFF
--- a/dosomething-dev/fastly-backend/main.tf
+++ b/dosomething-dev/fastly-backend/main.tf
@@ -71,9 +71,9 @@ resource "fastly_service_v1" "backends-dev" {
   }
 
   cache_setting {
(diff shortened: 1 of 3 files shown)


DFurnes:
Applied to development & QA. It's a little hard to test this works since our app code doesn't send cache headers for 404'd responses... but doesn't seem to have an adverse effect! 😅

DFurnes:
Applied! Merging in.

DFurnes:
References [DoSomething/internal#480](https://github.com/DoSomething/internal/issues/480).

PR #210, opened 2019-10-02
Define a log stream so this isn't a consistent "drift".

DFurnes:
This should clear up some "configuration drift" that was showing up [on `master`](https://app.terraform.io/app/dosomething/workspaces/infrastructure/runs/run-M2rC4ezVaPEzyymt), since Terraform wouldn't know about the `cloudwatch_log_stream` that was automatically created by AWS & try to reset it to the `null` default value.

So that we don't keep trying to delete this log stream on future deploys, we'll just make a single "quasar-vpn" log stream that Terraform can manage itself. 🎋

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@ This is DoSomething.org's infrastructure as code, built using [Terraform](https:
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


DFurnes:
This was also just a good excuse to try out remote plans & applies (see the little "check" on the PR/commit) because how cool is that?! 😸 No more need to manually copy-paste plans into PRs!

PR #213, opened 2019-10-16
Change default `acl` and remove `force_public` for S3 buckets.

DFurnes:
This pull request changes the default `acl` for S3 Buckets to `private`, and removes the potentially dangerous `force_public` option now that Rogue [no longer needs it](https://github.com/DoSomething/rogue/pull/918). This change will make objects uploaded into Rogue and Longshot's S3 buckets private by default.

References [#165980718](https://www.pivotaltracker.com/story/show/165980718)


DFurnes:
The plan has a fair amount of noise from "re-creating" some of the existing IAM bucket policies in cases where we're modifying the bucket's `acl` but the end result should be a no-op (aside from removing the "force public" policies from the three Rogue buckets). 🚳

sheyd:
sixdotsix

DFurnes:
🤘
 
> One random note the output reminded me of @DFurnes is should we put in a deprecation ticket for the `longshot` buckets once we're ready to deprecate the app?

Yes!! We had that in our Q4 planning doc, but not ticketized yet.

PR #214, opened 2019-11-07
Adds SCHOOLS_DB_URL env for GraphQL

aaronschachter:
And corresponding `greatschools_database_url` AWS SSM parameter per greatschools_database_url


DFurnes:
> ```
> ~ size     = "Standard-1X" -> "Hobby"
> ```

I'm curious why the plan shows it rebuilding some IAM policies & trying to resize some of our non-prod Heroku dynos down to the Hobby scale, but it's definitely not a change introduced here.

PR #218, opened 2020-01-14
Upping Quasar QA storage since it ran out.

sheyd:
### What's this PR do?

We ran out of storage space on Quasar QA while attempting a [refresh from Prod](https://www.pivotaltracker.com/story/show/164447976). Upping the storage to match Production values.

We'll see if it can upgrade the storage while in `storage-full` status. No other troubleshooting has been successful since we can't even connect to the server. 

...

### Relevant tickets

References [Pivotal #170705217](https://www.pivotaltracker.com/story/show/170705217).

### Checklist

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -445,7 +445,7 @@ data "aws_ssm_parameter" "prod_password" {
 }
 
 resource "aws_db_instance" "quasar-qa" {
-  allocated_storage               = 1000
+  allocated_storage               = 4000
   engine                          = "postgres"
   engine_version                  = "11.4"
   instance_class                  = "db.m5.2xlarge"


quasar/main.tf
```diff
@@ -445,7 +445,7 @@ data "aws_ssm_parameter" "prod_password" {
 }
 
 resource "aws_db_instance" "quasar-qa" {
-  allocated_storage               = 1000
+  allocated_storage               = 4000
```
DFurnes: How come this was set to a different value than production before?

DFurnes: It might also be worth looking into [storage auto-scaling](https://www.terraform.io/docs/providers/aws/r/db_instance.html#storage-autoscaling) in the future.

sheyd: We had smaller data sets in QA vs Prod so didn't justify the additional cost.

sheyd: > It might also be worth looking into [storage auto-scaling](https://www.terraform.io/docs/providers/aws/r/db_instance.html#storage-autoscaling) in the future.

OOOOO sweet, will add a backlog ticket!

DFurnes:
This change [will not take effect immediately](https://www.terraform.io/docs/providers/aws/r/db_instance.html):

> Changes to a DB instance can occur when you manually change a parameter, such as allocated_storage, and are reflected in the next maintenance window. Because of this, Terraform may report a difference in its planning phase because a modification has not yet taken place.

You can then apply the queued changes immediately via the AWS CLI:

```
aws rds modify-db-instance --db-instance-identifier quasar-qa --apply-immediately
```

sheyd:
@DFurnes Thanks for the reminder on this 😄 !

PR #220, opened 2020-01-16
Delete Rogue QA for fire drill.

DFurnes:
### What's this PR do?

This pull request deletes Rogue QA (Heroku app, SQS queue, IAM user, RDS database) so that we can test rebuilding it from scratch. We've `terraform state mv`'d the S3 bucket aside, since AWS won't let us delete it until the bucket is empty and that'd be a real slog.

👀

🔥 

### Relevant tickets

References [Pivotal #170080527](https://www.pivotaltracker.com/story/show/170080527).

### Checklist

### DIFF
--- a/dosomething-qa/main.tf
+++ b/dosomething-qa/main.tf
@@ -31,9 +31,9 @@ module "fastly-backend" {
   northstar_domain  = module.northstar.domain
   northstar_backend = module.northstar.backend
 
-  rogue_name    = module.rogue.name
-  rogue_domain  = module.rogue.domain
-  rogue_backend = module.rogue.backend
+  rogue_name    = "dosomething-rogue-qa"
+  rogue_domain  = "activity-qa.dosomething.org"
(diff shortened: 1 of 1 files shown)


dosomething-qa/main.tf
```diff
@@ -31,9 +31,9 @@ module "fastly-backend" {
   northstar_domain  = module.northstar.domain
   northstar_backend = module.northstar.backend
 
-  rogue_name    = module.rogue.name
-  rogue_domain  = module.rogue.domain
-  rogue_backend = module.rogue.backend
+  rogue_name    = "dosomething-rogue-qa"
+  rogue_domain  = "activity-qa.dosomething.org"
+  rogue_backend = "dosomething-rogue-qa.herokuapp.com"
```
DFurnes: We had to remove these references in order to avoid errors on Fastly QA. We should perhaps see if we can find a way to make these two modules less tightly coupled.

PR #225, opened 2020-01-23
Split into multiple Terraform workspaces.

DFurnes:
### What's this PR do?

This pull request splits our current "monolithic" Terraform state into separate workspaces, which lets us safely apply changes to environments selectively (for example, testing a change to a shared module in `dosomething-dev` before applying in production).

This splits our single `infrastructure` workspace into 9 independent workspaces:
 - `dosomething-prod`, `dosomething-qa`, and `dosomething-dev` for our core applications. In the future, these can be consolidated into [one single `dosomething` that we promote between workspace stages](https://www.terraform.io/docs/cloud/workspaces/repo-structure.html#multiple-workspaces-per-repo-recommended-) (and customize via environment variables in each stage).
- `longshot` (with our old Longshot application resources)
- `quasar` (which has Quasar & Quasar QA resources… should be split in the future!)
- `redirects` (with our two Fastly "domain redirect" properties)
- `voting-app` (with the old [Athletes Gone Good](http://www.athletesgonegood.com) static site)
- `vote-instapage` (with the Fastly wrapper around [`vote.dosomething.org`](https://vote.dosomething.org))
- `shared` (with Heroku pipelines & CircleCI credentials)

I have not updated the three `dosomething-` modules to import the pipeline variable from the `shared` pipeline yet, just to keep this a bit more contained (instead, these are just a variable for each workspace with the Pipeline ID).

Yikes!!! 🐳 This is not quite as overwhelming as it looks – the bulk of the diff comes from moving files that were referenced outside of the "module path" into each module where they're referenced (since you can't just grab files recklessly between workspaces anymore).

The other big change is moving backend & provider configuration into each module (e.g. from the single `main` top-level module into `dosomething`, `dosomething-qa`, etc). This is nice since it means we can test newer provider versions on dev/QA before using on production!

If you click to expand the "checks" at the bottom of this PR, you can see that this introduces no new changes:

Screen Shot 2020-01-24 at 2 59 48 PM

Moving resources between workspaces is incredibly awkward (hopefully improvements are on the Terraform roadmap!): you can either `terraform import` each one from scratch, or you can `terraform state pull` the state down to a file & `terraform state mv` each resource into a "fresh" state file that you can then `terraform state push` to the new workspace.

I opted for the second approach, but it was still way more tedious than I'd have expected!

### Relevant tickets

References [Pivotal #170776057](https://www.pivotaltracker.com/story/show/170776057).

### Checklist


dosomething-dev/fastly-backend/main.tf
```diff
@@ -144,14 +144,14 @@ resource "fastly_service_v1" "backends-dev" {
...
-    content           = file("${path.root}/shared/deny-robots.txt")
+    content           = file("${path.module}/deny-robots.txt")
     request_condition = "path-robots"
   }
 
   snippet {
     name    = "Shared - Set X-Origin-Name Header"
     type    = "fetch"
-    content = file("${path.root}/shared/app_name.vcl")
+    content = file("${path.module}/app_name.vcl")
```
DFurnes: Here's an example of why `app_name.vcl` now needs to be duplicated in each individual module – we can't (for security reasons, I assume) reach outside the "root" of the workspace's module to load files & so this needed to be swapped to `path.module` and moved within the `dosomething-dev/fastly-backend` directory.

dosomething-dev/main.tf
```diff
@@ -1,8 +1,56 @@
...
 variable "rogue_pipeline" {}
-variable "papertrail_destination" {}
-variable "papertrail_destination_fastly" {}
+
+terraform {
+  backend "remote" {
+    organization = "dosomething"
+
+    workspaces {
+      prefix = "dosomething-"
```
DFurnes: This could also easily be `name = "dosomething-dev"`, but this eases us into a future where we have a single `dosomething` module that gets applied to `-dev`, `-qa`, and `-prod` workspaces with different variables.

DFurnes: I'll tackle that in a follow-up ticket since this already got so much heftier than I'd imagined it would.

dosomething-qa/fastly-frontend/takeovers/election.html
```diff
@@ -0,0 +1,261 @@
+<!DOCTYPE html>
```
DFurnes: We should maybe just remove this (an old static HTML takeover from the election)! But didn't want to make any changes in this PR that'd result in an actual configuration change.

weerd: Yeah... is there a ticket regarding removing it? If not we should defs add one!

DFurnes: I need to make one! 📝

DFurnes: Created [#170936349](https://www.pivotaltracker.com/story/show/170936349) for this.

PR #237, opened 2020-04-03
Updating Quasar QA RDS instance size and associated parameters.

sheyd:
### What's this PR do?

This pull request:
* Reduces the RDS instance class for the Quasar QA data warehouse. 
* Adjusts the relevant DB parameters using the [PG Tune](https://pgtune.leopard.in.ua/#/) tool.
* Bumps to the latest minor release version.

Will make sure the Terraform validation runs.

We're tuning down/saving money where we can on unnecessary infrastructure costs.

### Relevant tickets

References [Pivotal #172147854](https://www.pivotaltracker.com/story/show/172147854).

### Checklist

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -265,7 +265,7 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
   # Sets effective RAM available to PG Query planner before using disk.
   parameter {
     name  = "effective_cache_size"
-    value = "24000000"
+    value = "12000000"
   }
 
   # Recommended by PGTuner tool: https://pgtune.leopard.in.ua/#/
@@ -280,7 +280,7 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
   # Based on 50 connections.
   parameter {
     name  = "work_mem"
-    value = "20971"
+    value = "41943"
   }
 
   # Recommended to only allow SSL connections from clients.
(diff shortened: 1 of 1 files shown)


sheyd:
Terraform plan [ran](https://app.terraform.io/app/dosomething/workspaces/quasar/runs/run-Zj45EY4fmNgPLfvd) successfully. One note is that the DB parameters will be tuned  _before_ the RDS instance re-sizing takes place. However, since we're tuning the parameters down, not up, it shouldn't impact current job runs (i.e. DB shouldn't run out of resources due to parameter over-provisioning).

quasar/main.tf
```diff
@@ -280,7 +280,7 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
   # Based on 50 connections.
   parameter {
     name  = "work_mem"
-    value = "20971"
+    value = "41943"
```
DFurnes, line 283: Any ideas why this went up? I'd have thought it would be roughly cut in half.

sheyd: Yeah, I was curious about that as well, and re-ran the numbers on the current instance before this PR went in, and looks like we had a bad copy/paste from a configuration using double the `max_connections` we actually use. 😬

DFurnes: Yikes! Good to catch it now.

PR #238, opened 2020-04-09
Adding first parts of S3 setup for C.io data pipeline.

sheyd:
### What's this PR do?

This pull request adds an S3 bucket under the Quasar module to enable using [Customer.io's data export](https://customer.io/docs/data-warehouse-sync) feature. 

Will review setup with Furnes, since likely need to add more module code to define specific IAM user for C.io.

### Relevant tickets

References [Pivotal # 172111876](https://www.pivotaltracker.com/story/show/172111876).

### Checklist

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -504,7 +504,7 @@ resource "aws_db_instance" "quasar" {
 # Provide S3 Bucket for Customer.io data file exports.
 module "iam_user" {
   source = "../components/iam_app_user"
-  name   = var.name
+  name   = "quasar-cio"
 }
 
 module "storage" {
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -499,3 +499,19 @@ resource "aws_db_instance" "quasar" {
   performance_insights_enabled    = true
   enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
 }
+
+
+# Provide S3 Bucket for Customer.io data file exports.
+module "iam_user" {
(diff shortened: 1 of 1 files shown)


quasar/main.tf
```diff
@@ -499,3 +499,19 @@ resource "aws_db_instance" "quasar" {
   performance_insights_enabled    = true
   enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
 }
+
+
+# Provide S3 Bucket for Customer.io data file exports.
+module "iam_user" {
+  source = "../components/iam_app_user"
+  name   = var.name
```
DFurnes, line 507: The build (which can be found by clicking the :x: on the commit) is failing with this error:

```
Error: Reference to undeclared input variable

  on main.tf line 507, in module "iam_user":
 507:   name   = var.name

An input variable with the name "name" has not been declared. This variable
can be declared with a variable "name" {} block.
```

That's because you don't have a [variable](https://www.terraform.io/docs/configuration/locals.html) named `name` in this module (you'd see a `variable {}` block at the top if you did). You could define a [local value](https://www.terraform.io/docs/configuration/locals.html) so you could reference the same "name" for both of these resources, or just provide an inline string here. This will be used as the name of the S3 bucket & user:

```suggestion
  name   = "quasar-cio"
```

sheyd: Woohoo! Setting a local value worked! Thanks @DFurnes!

PR #247, opened 2020-05-26
Adding PostgreSQL 12 settings for Quasar QA.

sheyd:
### What's this PR do?

This pull request sets up PostgreSQL 12 parameter group for Quasar QA server.

Terraform Cloud build review.

### Relevant tickets

References [Pivotal #172151665](https://www.pivotaltracker.com/story/show/172151665).

### Checklist

### DIFF
--- a/quasar/main.tf
+++ b/quasar/main.tf
@@ -343,6 +343,108 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
   }
 }
 
+resource "aws_db_parameter_group" "quasar-qa-pg12" {
+  name   = "quasar-qa-pg12"
+  family = "postgres12"
+
+  # Required for DMS CDC Replication:
+  # https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.v10
+  parameter {
+    name         = "rds.logical_replication"
+    value        = "1"
+    apply_method = "pending-reboot"
+  }
+
+  # Required for DMS CDC Replication:
+  # https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.v10
+  parameter {
+    name         = "max_logical_replication_workers"
+    value        = "500"
+    apply_method = "pending-reboot"
+  }
+
+  # Recommended by PGTuner tool: https://pgtune.leopard.in.ua/#/
+  # Sets effective RAM available to PG Query planner before using disk.
+  parameter {
+    name  = "effective_cache_size"
+    value = "12000000"
+  }
+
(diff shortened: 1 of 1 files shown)


quasar/main.tf
```diff
@@ -343,6 +343,107 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
...
+    name         = "max_logical_replication_workers"
+    value        = "500"
+    apply_method = "pending-reboot"
+  }
+
+  # Recommended by PGTuner tool: https://pgtune.leopard.in.ua/#/
+  # Sets effective RAM available to PG Query planner before using disk.
+  parameter {
+    name  = "effective_cache_size"
+    value = "12000000"
```
DFurnes, line 370: It might be nice to re-write these to make units clearer (is it KB?) and how we got here... e.g:

```hcl
value = "${12 * 1000 * 1000}" # 12GB.
```

quasar/main.tf
```diff
@@ -343,6 +343,107 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
...
+  # Recommended to only allow SSL connections from clients.
+  parameter {
+    name  = "rds.force_ssl"
+    value = "1"
+  }
+
+  # Only log queries with a duration longer than this to get slow queries.
+  parameter {
+    name  = "log_min_duration_statement"
+    value = "1000"
```
DFurnes, line 396: Also curious what the units are here... is this seconds? If so, let's just add an inline comment:

```hcl
value = "1000" # 16 minutes
```

quasar/main.tf
```diff
@@ -343,6 +343,107 @@ resource "aws_db_parameter_group" "quasar-qa-pg11" {
...
+  parameter {
+    name  = "log_duration"
+    value = "0"
+  }
+
+  # Testing turning synchronization off to see if improves performance
+  parameter {
+    name  = "synchronous_commit"
+    value = "off"
+  }
```
DFurnes, line 414: Is this a new thing we're testing? If not, did we see an improvement before? If there's a ticket, let's link it here.

quasar/main.tf
sheyd: Ooh, love this! Yah, will definitely change this!

quasar/main.tf
sheyd: Millseconds, will document.

quasar/main.tf
sheyd: This was part of some performance work we tested out about a year ago, when investigating some performance gremlins in the machine. Tried to look for a ticket attached to this, but couldn't find it. We didn't see a measurable performance improvement either way, so happy to remove it. Thoughts?

DFurnes: Yeah, let's remove this if we didn't find any improvement.

PR #250, opened 2020-05-27
Adding Fivetran logging IAM policy and role

blisteringherb:
### What's this PR do?

This pull request sets up the IAM policy and role to enable Fivetran Cloudwatch logging. 

I don't have access to terraform.io, so this needs to be tested locally and I'll need the output ARN to put into the Fivetran setup. There's also an additional variable, `fivetran_cloudwatch_integration_external_id` that's necessary for this to run successfully. 
...

Here are the setup instructions from Fivetran -https://fivetran.com/docs/logs/cloudwatch/setup-guide
and an example of a Fivetran terraform setup https://github.com/tropos-io/tf-fivetran-cloudwatch

...

### Relevant tickets

References [Pivotal #172623778](https://www.pivotaltracker.com/story/show/172623778).

### Checklist


quasar/variables.tf
```diff
@@ -0,0 +1,15 @@
+/*
+ ******************************************************
+ * Copyright (C) 2019 tropos.io <team@tropos.io>
+ *
+ ******************************************************
+ */
```
DFurnes, line 6: Is this file copyrighted? ©️

quasar/variables.tf
```diff
@@ -0,0 +1,15 @@
...
+
+variable "fivetran_cloudwatch_integration_external_id" {
+  description = "External ID for the trust relationship between fivetran and AWS"
+}
+
+variable "aws_region" {
+  description = "The region where the resources will be deployed"
+  default = "eu-west-1"
+}
```
DFurnes, line 15: This variable seems to be unused. (We default to `us-east-1` in our provider [here](https://github.com/DoSomething/infrastructure/blob/5d59c6309a0cd661c5a0933838f970c6ba08bc0d/quasar/main.tf#L13)).

quasar/main.tf
```diff
@@ -615,3 +615,60 @@ module "storage" {
...
+# Fivetran logging settings based on https://fivetran.com/docs/logs/cloudwatch/setup-guide
+# and https://github.com/tropos-io/tf-fivetran-cloudwatch
+
+data "aws_iam_policy_document" "fivetran_cloudwatch_integration_assume_role" {
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::834469178297:root"]
```
DFurnes, line 628: This is Fivetran's AWS account ID, right? Can we add a comment on this line noting that?

quasar/main.tf
```diff
@@ -615,3 +615,60 @@ module "storage" {
...
+    }
+
+    condition {
+      test     = "StringEquals"
+      variable = "sts:ExternalId"
+
+      values = [
+        "${var.fivetran_cloudwatch_integration_external_id}",
+      ]
```
DFurnes, line 637: Since we're not doing any string interpolation here, the wrapping double-quotes can be removed. While these will technically work, they'll be flagged as "warnings" on our Terraform builds, like this:

```
Warning: Interpolation-only expressions are deprecated

  on main.tf line 659, in resource "aws_iam_policy" "fivetran_cloudwatch_integration":
 659:   policy = "${data.aws_iam_policy_document.fivetran_cloudwatch_integration.json}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.

(and 3 more similar warnings elsewhere)
```

```suggestion
      values = [
        var.fivetran_cloudwatch_integration_external_id,
      ]
```

quasar/main.tf
```diff
@@ -615,3 +615,60 @@ module "storage" {
...
+      "logs:CreateLogStream",
+      "logs:PutLogEvents"
+    ]
+    resources = ["*"]
+  }
+}
+
+resource "aws_iam_policy" "fivetran_cloudwatch_integration" {
+  name   = "FivetranCloudwatchIntegrationPolicy"
+  policy = "${data.aws_iam_policy_document.fivetran_cloudwatch_integration.json}"
```
DFurnes, line 659: Same here:

```suggestion
  policy = data.aws_iam_policy_document.fivetran_cloudwatch_integration.json
```

quasar/main.tf
```diff
@@ -615,3 +615,60 @@ module "storage" {
...
+}
+
+resource "aws_iam_role" "fivetran_cloudwatch_integration" {
+  name               = "FivetranCloudwatchIntegrationRole"
+  description        = "Role for Fivetran Cloudwatch Integration"
+  assume_role_policy = "${data.aws_iam_policy_document.fivetran_cloudwatch_integration_assume_role.json}"
```
DFurnes, line 665:
```suggestion
  assume_role_policy = data.aws_iam_policy_document.fivetran_cloudwatch_integration_assume_role.json
```

quasar/main.tf
```diff
@@ -615,3 +615,60 @@ module "storage" {
...
+  tags = {
+    Application = "Fivetran"
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "fivetran_aws_integration" {
+  role       = "${aws_iam_role.fivetran_cloudwatch_integration.name}"
+  policy_arn = "${aws_iam_policy.fivetran_cloudwatch_integration.arn}"
```
DFurnes, line 673:
```suggestion
  role       = aws_iam_role.fivetran_cloudwatch_integration.name
  policy_arn = aws_iam_policy.fivetran_cloudwatch_integration.arn
```

quasar/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "RoleARN" {
+  description = "AWS Role ARN to be used in Fivetran"
+  value       = "${aws_iam_role.fivetran_cloudwatch_integration.arn}"
```
DFurnes, line 3:
```suggestion
  value       = aws_iam_role.fivetran_cloudwatch_integration.arn
```

quasar/main.tf
```diff
@@ -615,3 +615,9 @@ module "storage" {
   acl        = "private"
   versioning = true
 }
+
+module "fivetran_logging" {
+  source = "../components/fivetran_cloudwatch_role"
+
```
DFurnes, line 621: This module requires a `fivetran_cloudwatch_integration_external_id ` argument (see `variables.tf` above):

```hcl
module "fivetran_logging" {
  source = "../components/fivetran_cloudwatch_role"
  fivetran_cloudwatch_integration_external_id = "..."
}
```

DFurnes:
@blisteringherb Just a heads up that we're moving around some resources in state in #251, and so you may see a lot of "X will be created, X will be destroyed" until we merge that and you rebase. Shouldn't block anything, but may make plans a bit more difficult to read. Sorry in advance!

blisteringherb:
@DFurnes I can't see the error for Terraform, would you mind relaying that so I can make some fixes?

DFurnes:
Oh totally! Here ya go:

```
Error: Variables not allowed

On ../components/fivetran_cloudwatch_role/variables.tf line 3: Variables may
not be used here.
```

I think you need to remove the `default = var.fivetran_cloudwatch_integration_external_id` line because you're creating a recursive loop where the variable is defaulting to itself.

PR #252, opened 2020-06-03
Replace 'acl' argument with boolean 'private', defaulting to true.

DFurnes:
### What's this PR do?

This pull request replaces the existing `acl` option on S3 buckets with a boolean `private` option, defaulting to `true`. If this option is set, it will also use S3's [Block Public Access](https://aws.amazon.com/s3/features/block-public-access/) feature to ensure we don't upload objects that are individually marked "public".

I'd recommend reviewing this [without whitespace](https://github.com/DoSomething/infrastructure/pull/252/files?w=1).

I'm pulling [#169216617](https://www.pivotaltracker.com/story/show/169216617) into this sprint because we keep getting warnings about these apps, and we'll want to keep this app's buckets but swap them to be safely private.

### Relevant tickets

References [Pivotal #173154671](https://www.pivotaltracker.com/story/show/173154671).

### Checklist

### DIFF
--- a/dosomething/main.tf
+++ b/dosomething/main.tf
@@ -145,13 +145,13 @@ module "phoenix_preview" {
 module "rogue_backup" {
   source = "../components/s3_bucket"
   providers = {
-    aws = "aws.west"
+    aws = aws.west
   }
 
   name       = "dosomething-rogue-backup"
-  acl        = "private"
   versioning = true
   archived   = true
+  private    = true
 }
 
 module "rogue" {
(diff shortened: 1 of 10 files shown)


dosomething/main.tf
```diff
@@ -145,13 +145,13 @@ module "phoenix_preview" {
 module "rogue_backup" {
   source = "../components/s3_bucket"
   providers = {
-    aws = "aws.west"
+    aws = aws.west
```
DFurnes, line 148: This is just to address a deprecation warning – this is a reference to the `aws.west` provider resource, not a string.

PR #254, opened 2020-06-03
Fix an AWS error encountered when applying #253.

DFurnes:
### What's this PR do?

This pull request (hopefully!) fixes [an issue](https://app.terraform.io/app/dosomething/workspaces/longshot/runs/run-ZcGBsivAvKwyS7Hg) applying #253, where we'd try to delete the IAM user before the attached policies had been deleted. According to [Terraform's documentation](https://www.terraform.io/docs/providers/aws/r/iam_user.html) and [this issue](https://git.io/JfPQF), adding `force_destroy` will let us delete this safely & sidestep this kaboom:

```
Error: Error deleting IAM User longshot-footlocker: DeleteConflict: Cannot delete entity, must delete policies first.
	status code: 409, request id: fc6a7002-5040-44da-a871-514d28a5d077
```

Step one is re-adding the IAM user so we can "apply" the `force_destroy` option, step two is trying to delete it again for real. Luckily, we can conditionally apply this per workspace (e.g. Longshot first, dev, QA, and finally prod) so we'll be able to carefully test before it hits the big leagues.

🙃 💥 

### Relevant tickets

References [Pivotal #169216617](https://www.pivotaltracker.com/story/show/169216617).

### Checklist

### DIFF
--- a/applications/longshot/main.tf
+++ b/applications/longshot/main.tf
@@ -59,6 +59,11 @@ module "database" {
   security_groups = ["sg-c9a37db2"]
 }
 
+module "iam_user" {
+  source = "../../components/iam_app_user"
+  name   = var.name
+}
+
 module "storage" {
   source  = "../../components/s3_bucket"
   name    = var.name
--- a/components/iam_app_user/main.tf
+++ b/components/iam_app_user/main.tf
@@ -4,6 +4,10 @@ variable "name" {
 
 resource "aws_iam_user" "aws_user" {
   name = var.name
+
+  # When destroying this resource, destroy even if it has non-Terraform
+  # managed access keys, policies, etc. See:
+  force_destroy = true
 }
 
 resource "aws_iam_access_key" "aws_key" {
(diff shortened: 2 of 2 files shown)


applications/longshot/main.tf
```diff
@@ -59,6 +59,11 @@ module "database" {
   security_groups = ["sg-c9a37db2"]
 }
 
+module "iam_user" {
+  source = "../../components/iam_app_user"
+  name   = var.name
+}
```
DFurnes, line 65: This is just reverting the deletion of these lines in #253.

components/iam_app_user/main.tf
```diff
@@ -4,6 +4,10 @@ variable "name" {
 
 resource "aws_iam_user" "aws_user" {
   name = var.name
+
+  # When destroying this resource, destroy even if it has non-Terraform
+  # managed access keys, policies, etc. See: https://git.io/JfP7O
+  force_destroy = true
```
DFurnes, line 10: Despite being named "force destroy", I do not believe this will force destruction on its own (it just forces it along if it already is going to destroy the resource for another reason). This is similar to a setting we have to use on our [Fastly properties](https://github.com/DoSomething/infrastructure/blob/fe6fac17c57d1179261a6333d2753483b389b9cf/dosomething/fastly-backend/main.tf#L23) to ensure that Terraform can delete them if it needs to.

I'll apply this change to Longshot, dev, and QA properties first to double-check that assumption! 🌵

PR #257, opened 2020-06-23
Add 'fivetran_s3_role' resource & configure for Bertly.

DFurnes:
### What's this PR do?

This pull request adds a [`fivetran_s3_role`](https://github.com/DoSomething/infrastructure/tree/fivetran-s3-role/components/fivetran_s3_role) module, which allows Fivetran to access buckets on our account.

This is based on the [Fivetran S3 Setup Guide](https://fivetran.com/docs/files/aws-s3/setup-guide) here, and follows the example set by @blisteringherb's work in #250 (to add a similar connection for CloudWatch logs). Since we want to give Fivetran access to only a few specific buckets, I've opted to add this as a resource that we attach to any buckets that need it (rather than a standalone "global" role).

I'll try creating a connector on development & QA (which have relatively few clicks) before doing so on production.

This will allow us to replicate this private S3 bucket (where Bertly 2.0 stores click information) into our data warehouse with Fivetran. We're storing clicks here, instead of PostgreSQL, to save money & hopefully reduce complexity.

### Relevant tickets

References [Pivotal #172652716](https://www.pivotaltracker.com/story/show/172652716).

### Checklist

### DIFF
--- a/components/fivetran_s3_role/main.tf
+++ b/components/fivetran_s3_role/main.tf
@@ -1,0 +1,19 @@
+data "aws_ssm_parameter" "external_id" {
+  name = "/fivetran/${var.environment}/external-id"
+}
+
+resource "aws_iam_policy" "fivetran_policy" {
+  name = "${var.name}-s3-fivetran"
+  policy = templatefile("${path.module}/iam-policy.json.tpl", {
+    bucket_arn = var.bucket.arn
+  })
+}
+
+resource "aws_iam_role" "fivetran_role" {
+  name = "${var.name}-s3-fivetran"
+
+  assume_role_policy = templatefile("${path.module}/iam-role.json.tpl", {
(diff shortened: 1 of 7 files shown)


components/fivetran_s3_role/main.tf
```diff
@@ -0,0 +1,19 @@
...
+  policy = templatefile("${path.module}/iam-policy.json.tpl", {
+    bucket_arn = var.bucket.arn
+  })
+}
+
+resource "aws_iam_role" "fivetran_role" {
+  name = "${var.name}-s3-fivetran"
+
+  assume_role_policy = templatefile("${path.module}/iam-role.json.tpl", {
+    fivetran_account_id = 834469178297
```
weerd, line 16: Just want to make sure it's fine to not have in env var ;)

PR #260, opened 2020-07-06
Add additional tags to AWS resources.

DFurnes:
### What's this PR do?

This pull request adds additional "tags" to AWS resources. These can be used to filter items in the web console (e.g. "show me all production databases") or CLI, or for analyzing our bill to see breakdown & trends by app.

I decided to start by categorizing our resources by:
- `Application` (e.g. `dosomething-northstar`)
- `Environment` (e.g. `development` or `production`)
- `Stack` (e.g. `web`, `backend`, or `data`).

I'm open to better naming for that last one – the idea is to have a way to group different "domain contexts" so we can group different parts of our stack. I considered `Context` but it seemed nebulous, and [`Domain`](https://en.wikipedia.org/wiki/Domain-driven_design) but figured it could be confused with a web domain.

And we can always adjust or add more tags in the future!

### How should this be reviewed?

👀

### Any background context you want to provide?

I know the diff looks scary, but it's all very straightforward changes!

### Relevant tickets

References [Pivotal #173665994](https://www.pivotaltracker.com/story/show/173665994).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.

PR #261, opened 2020-07-06
Fix some issues with new AWS tags.

DFurnes:
### What's this PR do?

This pull request fixes some issues I noticed after applying #260 and reviewing resources in the AWS console:

📄 Fixes some unclear variable descriptions. 1343e3e

🚀 Adds tags to the "deploy buckets" that are used to deploy Lambda functions (previously untagged). 488cb3b

📛 Some S3, SQS, and Lambda resources have suffixes (like `dosomething-northstar-low` and `dosomething-northstar-high` queues). I've added a new `application` variable so these can be grouped as `dosomething-northstar`. 914c2c9

🏹 I've grouped all the (now deprecated) Longshot S3 buckets under a `dosomething-longshot` Application tag, rather than separate `Application=longshot-hrblock`, `Application=longshot-footlocker`, etc. 87ebde6

👀

This is useful for reviewing AWS resources in the admin panel or billing views!

### Relevant tickets

References [Pivotal #173665994](https://www.pivotaltracker.com/story/show/173665994).

### Checklist

### DIFF
--- a/applications/bertly/main.tf
+++ b/applications/bertly/main.tf
@@ -54,6 +54,7 @@ locals {
 module "app" {
   source = "../../components/lambda_function"
 
+  application = var.name
   name        = var.name
   environment = var.environment
   stack       = local.stack
@@ -93,6 +94,7 @@ module "database" {
 module "storage" {
   source = "../../components/s3_bucket"
 
+  application = var.name
   name        = "${var.name}-logs"
   environment = var.environment
   stack       = local.stack
--- a/applications/graphql/main.tf
+++ b/applications/graphql/main.tf
@@ -73,6 +73,7 @@ locals {
 module "app" {
   source = "../../components/lambda_function"
 
+  application = var.name
   name        = var.name
   environment = var.environment
   stack       = local.stack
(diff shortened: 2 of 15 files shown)


katiecrane:
So is the `name` like "Rogue" but the `application` would be like "dosomething-rogue"?

DFurnes:
Yeah, I'm wondering if there's a better name (hah) for `name` now... here's some examples:

| `name` | `application` | description
| ------- | --------------| -----------|
| dosomething-northstar-low | dosomething-northstar | Northstar's low-priority SQS queue.
| dosomething-northstar-high | dosomething-northstar | Northstar's high-priority SQS queue.
| dosomething-graphql | dosomething-graphql | Our GraphQL Lambda function.
| dosomething-graphql-webhook | dosomething-graphql | Our GraphQL Contentful webhook Lambda function.

So `name` is the specific _thing_ (since sometimes applications will have more than one bucket, queue, or function), and `application` is the same for all resources in a given app, no matter how many there are.

katiecrane:
Hmmm okay yeah I think I'm confused because in Heroku we have "pipeline" and "app" and here we have "name" and "application" but the AWS "name" corresponds to the Heroku "app" 😵 But that table is very helpful and given that I think this is good to go!

DFurnes:
Oh – those are good points! I'd definitely be down for re-considering all this naming to be clearer across the board.

katiecrane:
Would love to help think on that! Nothing super clear came to mind immediately (since "pipeline" doesn't really apply)

DFurnes:
Let's keep mulling on it!

PR #263, opened 2020-07-20
Add Papertrail archive bucket.

DFurnes:
### What's this PR do?

This pull request adds an S3 bucket where we can keep our own archive of our Papertrail logs. This is could be useful if we ever need to search further than the four-week window included in our plan (via [Athena](https://aws.amazon.com/athena/)), or if we ever need to reference logs older than a year (say, to ensure a breach didn't occur if we found a security hole).

This sets up a new bucket to store these zipped archive files, roughly following [Papertrail's setup guide](https://help.papertrailapp.com/kb/how-it-works/automatic-s3-archive-export/). This bucket should be private & only allow write access to Papertrail's AWS account ID.

🎋

### Relevant tickets

References [Pivotal #173253029](https://www.pivotaltracker.com/story/show/173253029).

### Checklist

### DIFF
--- a/components/log_archive/main.tf
+++ b/components/log_archive/main.tf
@@ -1,0 +1,27 @@
+variable "name" {
+  description = "The name for the archive bucket (e.g. dosomething-papertrail)."
+}
+
+resource "aws_s3_bucket" "archive" {
+  bucket = var.name
+  acl    = "private"
+
+  tags = {
+    Application = "papertrail"
+    Environment = "all"
+    Stack       = "all"
+  }
+}
+
+resource "aws_s3_bucket_policy" "papertrail_access_policy" {
+  bucket = aws_s3_bucket.archive.id
+  policy = templatefile("${path.module}/policy.json.tpl", { bucket_arn = aws_s3_bucket.archive.arn })
+}
+
+resource "aws_s3_bucket_public_access_block" "private_policy" {
+  bucket = aws_s3_bucket.archive.id
+
+  block_public_acls   = true
+  block_public_policy = true
+  ignore_public_acls  = true
+}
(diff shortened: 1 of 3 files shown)


components/log_archive/main.tf
```diff
@@ -0,0 +1,27 @@
...
+}
+
+resource "aws_s3_bucket" "archive" {
+  bucket = var.name
+  acl    = "private"
+
+  tags = {
+    Application = "papertrail"
+    Environment = "all"
+    Stack       = "all"
```
DFurnes, line 12: I wasn't sure what to put here – since Papertrail is sort-of a "catch all". Open to suggestions!

components/log_archive/policy.json.tpl
```diff
@@ -0,0 +1,21 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Sid": "PapertrailLogArchive",
+        "Effect": "Allow",
+        "Principal": {
+          "AWS": [
+            "arn:aws:iam::719734659904:root"
```
DFurnes, line 9: This is Papertrail's AWS account ID, referenced from [this documentation](https://help.papertrailapp.com/kb/how-it-works/automatic-s3-archive-export/#alternative-define-sharing-policy-with-iam).

components/log_archive/main.tf
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+resource "aws_s3_bucket_policy" "papertrail_access_policy" {
+  bucket = aws_s3_bucket.archive.id
+  policy = templatefile("${path.module}/policy.json.tpl", { bucket_arn = aws_s3_bucket.archive.arn })
```
katiecrane, line 18: So this line and the `policy.json.tpl` file are saying that the Papertrail AWS user can put and delete things in this bucket?

DFurnes: Yep!

PR #266, opened 2020-08-03
Use 'dosome.click' for Bertly 2.0

DFurnes:
### What's this PR do?

This pull request updates the domain mapping for Bertly 2.0 to use dosome.click.

:eyes:

It's launch day! :rocket: 

### Relevant tickets

References [Pivotal #172652963](https://www.pivotaltracker.com/story/show/172652963).

### Checklist

### DIFF
--- a/dosomething/main.tf
+++ b/dosomething/main.tf
@@ -69,8 +69,8 @@ module "bertly" {
 
   environment   = "production"
   name          = "dosomething-bertly"
-  domain        = "new.dosome.click"
-  certificate   = "*.dosome.click"
+  domain        = "dosome.click"
+  certificate   = "dosome.click"
   northstar_url = "https://identity.dosomething.org"
   logger        = module.papertrail
 }


dosomething/main.tf
```diff
@@ -69,8 +69,8 @@ module "bertly" {
 
   environment   = "production"
   name          = "dosomething-bertly"
-  domain        = "new.dosome.click"
-  certificate   = "*.dosome.click"
+  domain        = "dosome.click"
+  certificate   = "dosome.click"
```
katiecrane, line 73: Is the old Bertly still somewhere using `dosome.click`?

DFurnes: Yep! I'm going to detach that mapping before I apply this. :link:

katiecrane: Ah so that one is outside of terraform?

katiecrane: Sounds good though - now I'm just curious!

DFurnes: Yes! Originally, Bertly was managed with [CloudFormation](https://aws.amazon.com/cloudformation/), Amazon's configuration-as-code tool. It had some nice integrations in the AWS web GUI, but also only worked with AWS services which was a deal-breaker for a lot of our stuff.

PR #269, opened 2020-09-03
Remove unused Longshot resources.

DFurnes:
### What's this PR do?

This pull request removes the now unused "redirect" applications we'd previously used to send traffic to these domains to SurveyMonkey Apply. These domains now either route through our nginx box or directly to Squarespace.

I fixed some unrelated formatting warnings in d112371, and removed these resources in 7a412c8.

🔥

### Relevant tickets

References [Pivotal #174579717](https://www.pivotaltracker.com/story/show/174579717).

### Checklist

### DIFF
--- a/longshot/main.tf
+++ b/longshot/main.tf
@@ -37,49 +37,23 @@ provider "null" {
   version = "~> 2.1"
 }
 
-resource "heroku_pipeline" "longshot" {
-  name = "longshot"
-}
-
 module "longshot-footlocker" {
   source = "../applications/longshot"
 
   name        = "longshot-footlocker"
-  domain      = "footlockerscholarathletes.com"
-  pipeline    = "${heroku_pipeline.longshot.id}"
   environment = "production"
-  deprecated  = true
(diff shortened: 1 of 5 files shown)


longshot/main.tf
```diff
@@ -37,49 +37,23 @@ provider "null" {
...
-}
-
 module "longshot-footlocker" {
   source = "../applications/longshot"
 
   name        = "longshot-footlocker"
-  domain      = "footlockerscholarathletes.com"
-  pipeline    = "${heroku_pipeline.longshot.id}"
   environment = "production"
-  deprecated  = true
```
katiecrane, line 51: Do we still want this deprecated flag?

DFurnes: It had originally been a way to tear down the database (since you first need to delete users & disable "deletion protection" by "deprecating" that resource before actually deleting it). At this point, it's no longer being used.

katiecrane (approved):
makes sense to me!

PR #270, opened 2020-09-17
Clean up Fastly backend modules.

DFurnes:
### What's this PR do?

This pull request cleans up our Fastly modules a bit, and extracts a shared `fastly_backends` component that we can use to create these Fastly properties in all three environments (development, QA, and production).

I've done a little miscellaneous cleanup in 4fc0a59 and extracted the shared module in d92bf15. A lot of the "noise" in this pull request comes from deleting the old duplicated (per environment) Terraform configs and snippets.

I'm doing this work now because I'm going to be making changes to our Fastly properties in [#173719273](https://www.pivotaltracker.com/story/show/173719273), and 

### Relevant tickets

References [#165498108](https://www.pivotaltracker.com/story/show/165498108) and [#170775632](https://www.pivotaltracker.com/story/show/170775632).

### Checklist


dosomething/fastly-frontend/main.tf
```diff
@@ -37,18 +37,6 @@ resource "fastly_service_v1" "frontend" {
...
-    name      = "timed-synthetic-takeover"
-    statement = "req.http.X-Timed-Synthetic-Response == \"true\""
-  }
-
-  response_object {
-    name              = "timed synthetic takeover"
-    request_condition = "timed-synthetic-takeover"
-    content           = file("${path.module}/takeovers/election.html")
-  }
-
```
DFurnes, line 51: This is from an old homepage takeover. 

It's something that we could re-add (from Git history) if we were to run again.

dosomething/fastly-frontend/main.tf
```diff
@@ -149,51 +137,51 @@ resource "fastly_service_v1" "frontend" {
...
   snippet {
-    name    = "Frontend - ISO-3166-2 Response Header"
+    name    = "ISO-3166-2 Response Header"
     type    = "deliver"
     content = file("${path.module}/iso3166_deliver.vcl")
   }
 
   snippet {
-    name    = "Frontend - Trigger International Redirect"
+    name    = "Trigger Homepage Redirect"
```
DFurnes, line 152: These were all just renamed for clarity.

components/fastly_backends/main.tf
```diff
@@ -1,11 +1,17 @@
...
-variable "papertrail_log_format" {}
+variable "name" {
+  description = "The name for this Fastly property."
+  type        = string
+}
+
+variable "backends" {
+  description = "The domains/backends to route traffic to in this property."
+  type        = list(object({ name = string, domain = string, backend = string }))
+}
```
DFurnes, line 9: This allows us to pass "backend" modules as long as they have `name`, `domain`, and `backend` output variables! This way we don't need to define individual `northstar_name`, `northstar_domain`, `northstar_backend`, `rogue_name`, etc. but can just pass in a list of backends like `module.northstar, module.rogue`.

components/fastly_backends/main.tf
```diff
@@ -15,48 +21,76 @@ locals {
   }
 }
 
-resource "fastly_service_v1" "backends-qa" {
-  name          = "Terraform: Backends (QA)"
+resource "fastly_service_v1" "backends" {
+  name          = "Terraform: Backends"
```
DFurnes: Whoops - fixed this not to be hard-coded in 8b21e4f.

components/fastly_backends/main.tf
```diff
@@ -15,48 +21,76 @@ locals {
...
+      address           = backend.value.backend
+      name              = backend.value.name
+      request_condition = "backend-${backend.value.name}" # defined above.
+      port              = 443
+
+      # We use shielding to increase chances that we get cache hits. We don't rely
+      # on Fastly for load balancing (since that is handled by Heroku's router).
+      shield           = "bwi-va-us"
+      auto_loadbalance = false
+    }
```
DFurnes, line 70: It's awkward to read with how the diff is broken up, but this basically replaces individual `domain { … }`, `condition { … }` and `backend { … }` blocks for each application with a "dynamic" block that builds all of these based on the provided `var.backends` variable.

components/fastly_backends/main.tf
```diff
@@ -15,48 +21,76 @@ locals {
...
 
-  condition {
-    type      = "RESPONSE"
-    name      = "response-rogue-qa"
-    statement = "req.http.host == \"${var.rogue_domain}\""
+  # Fix issue with geolocation when shielding is enabled:
+  snippet {
+    name    = "Fix Geolocation With Shielding On"
+    type    = "recv"
+    content = file("${path.module}/recv_geolocation.vcl")
```
DFurnes, line 77: I moved up this from the bottom of the file since it's relevant to the `shield` arg on the backend above.

components/fastly_backends/main.tf
```diff
@@ -15,48 +21,76 @@ locals {
...
     type      = "REQUEST"
     name      = "path-robots"
     statement = "req.url.basename == \"robots.txt\""
   }
 
+  response_object {
+    name              = "robots.txt deny"
+    content           = file("${path.module}/deny-robots.txt")
+    request_condition = "path-robots"
+  }
```
DFurnes, line 91: Likewise, moved the `response_object` here up so it's next to the condition it relies on.

components/fastly_backends/main.tf
```diff
@@ -152,34 +169,24 @@ resource "fastly_service_v1" "backends-qa" {
...
-    content = file("${path.module}/recv_geolocation.vcl")
-  }
-
+  # Write requests to this Fastly property to the provided Papertrail drain:
   papertrail {
     name    = "backend"
     address = element(split(":", var.papertrail_destination), 0)
     port    = element(split(":", var.papertrail_destination), 1)
-    format  = var.papertrail_log_format
+    format  = "%t '%r' status=%>s app=%%{X-Application-Name}o cache=\"%%{X-Cache}o\" country=%%{X-Fastly-Country-Code}o ip=\"%a\" user-agent=\"%%{User-Agent}i\" service=%%{time.elapsed.msec}Vms"
```
DFurnes, line 190: Since we're now re-using the same `fastly_backends` service across all environments, I don't think it's as useful to pass in the `papertrail_log_format` variable here like we did before.

katiecrane: @DFurnes so this string is dynamically building what would have previously been saved as `var.papertrail_log_format`?

DFurnes: Yep, I've just directly copied in the string value we used to send for this (as `var.papertrail_log_format`) from `dosomething/main.tf`, `dosomething-qa/main.tf`, and `dosomething-dev/main.tf`. I think it makes more sense to keep here now, since we can define it one place for all three environments.

DFurnes: Here's, for example, where we used to grab this [for the production environment](https://github.com/DoSomething/infrastructure/blob/13f1ec7367067a89746f605e75704aaa95eb397d/dosomething/main.tf#L56).

dosomething-dev/main.tf
```diff
@@ -78,22 +78,15 @@ module "fastly-frontend" {
...
-  phoenix_domain  = module.phoenix.domain
-  phoenix_backend = module.phoenix.backend
-
-  rogue_name    = module.rogue.name
-  rogue_domain  = module.rogue.domain
-  rogue_backend = module.rogue.backend
+  backends = [
+    module.northstar,
+    module.rogue,
+  ]
```
katiecrane, line 87: So here is where we leverage [this work](https://github.com/DoSomething/infrastructure/pull/270/files#r490504424)?

DFurnes: Yes! We send in the set of backend modules each place we use this (so e.g. `identity-dev.dosomething.org` and `activity-dev.dosomething.org`'s variables would be passed in here). We read this provided variable on the line you linked, and then iterate over that array to configure the Fastly property below.

components/fastly_backends/main.tf
katiecrane: Makes sense to me, love having this long thing just in one place!

katiecrane (approved):
This was very easy to review because of the super helpful comments you added, thank you Dave!

PR #271, opened 2020-09-18
Create Fastly property as part of "static" app module.

DFurnes:
### What's this PR do?

This pull request extracts [assets.dosomething.org](https://assets.dosomething.org) into its own Fastly property, inside the `static` application module. This makes it easier to then re-use a single "frontend" module between all three environments (since we only create [assets.dosomething.org](https://assets.dosomething.org) in production), and simplifies our production Fastly config (to reduce the chance of accidental errors)!

👀

I'm going to follow this up with a similar PR to #270 which creates a single re-usable Fastly front-end module (for [#165498108](https://www.pivotaltracker.com/story/show/165498108)), and that in turn will make it nice and simple to add security headers for [#173719273](https://www.pivotaltracker.com/story/show/173719273).

### Relevant tickets

References [Pivotal #174871199](https://www.pivotaltracker.com/story/show/174871199).

### Checklist

### DIFF
--- a/applications/static/main.tf
+++ b/applications/static/main.tf
@@ -10,6 +10,64 @@ variable "stack" {
   description = "The 'stack' for this bucket: web, sms, backend, data."
 }
 
+resource "fastly_service_v1" "cdn" {
+  name          = "Terraform: ${var.domain}"
+  force_destroy = true
+
+  domain {
+    name = var.domain
+  }
+
+  backend {
+    name    = "s3-${var.domain}"
+    address = "${aws_s3_bucket.bucket.id}.s3-website-${aws_s3_bucket.bucket.region}.amazonaws.com"
+    port    = 80
+  }
+
+  request_setting {
+    name      = "Force SSL"
+    force_ssl = true
+  }
+
+  gzip {
+    name = "gzip"
+
+    extensions = ["css", "js", "html", "eot", "ico", "otf", "ttf", "json"]
+
+    content_types = [
+      "text/html",
+      "application/x-javascript",
+      "text/css",
+      "application/javascript",
+      "text/javascript",
+      "application/json",
+      "application/vnd.ms-fontobject",
+      "application/x-font-opentype",
+      "application/x-font-truetype",
+      "application/x-font-ttf",
+      "application/xml",
+      "font/eot",
+      "font/opentype",
+      "font/otf",
+      "image/svg+xml",
+      "image/vnd.microsoft.icon",
+      "text/plain",
+      "text/xml",
+    ]
+  }
+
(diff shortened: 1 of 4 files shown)


applications/static/main.tf
```diff
@@ -10,6 +10,64 @@ variable "stack" {
...
+  # expected 'Vary' if it isn't already set on the response.
+  header {
+    name        = "S3 Vary"
+    type        = "cache"
+    action      = "set"
+    destination = "http.Vary"
+    source      = "\"Origin, Access-Control-Request-Headers, Access-Control-Request-Method\""
+  }
+}
+
```
DFurnes, line 70: This is extracted from the production `fastly-frontend` module (below). Since we now only handle one domain in each of these two properties, we no longer need request conditions for backends/headers.

applications/static/main.tf
```diff
@@ -66,12 +124,3 @@ resource "aws_s3_bucket_policy" "bucket_policy" {
...
 }
-
-output "domain" {
-  value = var.domain
-}
-
-output "backend" {
-  value = "${aws_s3_bucket.bucket.id}.s3-website-${aws_s3_bucket.bucket.region}.amazonaws.com"
-}
-
```
DFurnes, line 77: Since we consume the S3 bucket "backend" inside the module, it no longer makes sense to output these.

voting-app/main.tf
```diff
@@ -21,48 +21,6 @@ provider "fastly" {
...
-    ]
-  }
-
-  backend {
-    name    = "s3-www.athletesgonegood.com"
-    address = module.agg.backend
-    port    = 80
-  }
-}
-
```
DFurnes, line 65: This now exists within the `agg` instance of the `static` module, below.

applications/static/main.tf
katiecrane: I don't see anything about `assets.dosomething.org` specifically referenced here like I do in `fastly-frontend` down below, can you explain this a little more?

DFurnes: Yeah, totes! We read the domain that the Fastly property & S3 bucket will be configured for using the `var.domain` variable in this module (declared at the top of this file, and used in the Fastly `domain { … }` block & bucket config).

When using this module to create resources in `voting-app/main.tf` and `dosomething/main.tf`, we can then pass a different domain (either `athletesgonegood.com` or `assets.dosomething.org`) to customize each "instance".

katiecrane: Ah I think I see now, and I do see those being passing through elsewhere!

katiecrane (approved):
good from me!

PR #272, opened 2020-09-18
Don't force SSL on Athletes Gone Good.

DFurnes:
### What's this PR do?

This pull request adds a toggle for whether or not to force SSL on a `static` application. This is a quick follow-up from #271 when I realized that we don't have an SSL certificate for [athletesgonegood.com](http://www.athletesgonegood.com) and so we can't force SSL there.

👀

In general, we of course want to force SSL! However, some old domains we keep around for historical purposes, like [athletesgonegood.com](http://www.athletesgonegood.com) aren't worth paying Fastly for a certificate for.

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/applications/static/main.tf
+++ b/applications/static/main.tf
@@ -10,6 +10,11 @@ variable "stack" {
   description = "The 'stack' for this bucket: web, sms, backend, data."
 }
 
+variable "force_ssl" {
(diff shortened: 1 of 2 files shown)


katiecrane (approved):
good catch!

PR #273, opened 2020-09-18
Extract shared Fastly frontend module.

DFurnes:
### What's this PR do?

This pull request extracts a shared `fastly-frontend` module that we can use to configure our www.dosomething.org, qa.dosomething.org, and dev.dosomething.org using the same Terraform config! This means we no longer have to copy-paste a change between all three environments, and reduces the chance that environments get out of sync.

This is the last of these hefty PRs, I promise! I'll drop notes in the diff to explain changes, but for the most part the bulk are either lines that were deleted or simply moved to a new folder. 🚚 

This makes the next set of functional changes I need to make to these Fastly properties a lot easier!

### Relevant tickets

References [Pivotal #165498108](https://www.pivotaltracker.com/story/show/165498108).

### Checklist


components/fastly_backends/main.tf
```diff
@@ -3,8 +3,8 @@ variable "name" {
   type        = string
 }
 
-variable "backends" {
-  description = "The domains/backends to route traffic to in this property."
+variable "applications" {
+  description = "The application modules to route traffic to."
```
DFurnes, line 7: I renamed the `backends` argument provided to `applications` for two reasons:

1. It's odd to provide a "backends" object that itself is composed of name, domain, and backend(!) properties.
2. I wanted to follow the same convention (`applications = […]` and `application = …`) for both these Fastly modules! It would feel somewhat awkward to pass a `backend` argument to the frontend module! 😂

components/fastly_frontend/main.tf
```diff
@@ -1,7 +1,17 @@
...
-variable "papertrail_log_format" {}
+variable "name" {
+  description = "The name for this Fastly property."
+  type        = string
+}
+
+variable "application" {
+  description = "The application module to route traffic to."
+  type        = object({ name = string, domain = string, backend = string })
+}
```
DFurnes, line 9: Following the same pattern established in #270, you can just pass in an application module here (as long as it has these named outputs). So we can just say `application = module.phoenix` for example, rather than needing to pass each piece separately! 😌

components/fastly_frontend/main.tf
```diff
@@ -159,7 +169,7 @@ resource "fastly_service_v1" "frontend" {
     name    = "frontend"
     address = element(split(":", var.papertrail_destination), 0)
     port    = element(split(":", var.papertrail_destination), 1)
-    format  = var.papertrail_log_format
+    format  = "%t '%r' status=%>s app=%%{X-Application-Name}o cache=\"%%{X-Cache}o\" country=%%{X-Fastly-Country-Code}o ip=\"%a\" user-agent=\"%%{User-Agent}i\" service=%%{time.elapsed.msec}Vms"
```
DFurnes, line 172: Like in #270, I've inlined this here since we always want the same log format regardless of what environment we're in. (We may, however, eventually want different log formats between frontend & backend properties).

dosomething-dev/main.tf
```diff
@@ -52,10 +52,6 @@ provider "null" {
   version = "~> 2.1"
 }
 
-locals {
-  papertrail_log_format = "%t '%r' status=%>s app=%%{X-Application-Name}o cache=\"%%{X-Cache}o\" country=%%{X-Fastly-Country-Code}o ip=\"%a\" user-agent=\"%%{User-Agent}i\" service=%%{time.elapsed.msec}Vms"
-}
```
DFurnes, line 57: After inlining this into the backend module in #270 and the frontend module here, this is no longer used!

dosomething-dev/main.tf
```diff
@@ -68,20 +64,18 @@ module "bertly" {
...
 
+  application            = module.phoenix
   papertrail_destination = var.papertrail_destination_fastly
-  papertrail_log_format  = local.papertrail_log_format
 }
 
 module "fastly-backend" {
   source = "../components/fastly_backends"
-  name   = "Fastly: Backends (Development)"
+  name   = "Terraform: Backends (Development)"
```
DFurnes, line 76: I'd accidentally prefixed these `Fastly:` in my previous PR... which, duh! Of course they're Fastly! I'm not sure how important it is anymore, but the reason I'd prefixed these with `Terraform:` originally was to indicate to anyone logged into the web console that they should make edits there (not via the GUI)!

components/fastly_frontend/main.tf
```diff
@@ -155,11 +169,29 @@ resource "fastly_service_v1" "frontend" {
...
+  # Configure 'robots.txt' global deny for backends:
+  condition {
+    type      = "REQUEST"
+    name      = "path-robots"
+    statement = "req.url.basename == \"robots.txt\""
+  }
+
+  dynamic "response_object" {
+    # If we're not in production, add a 'robots.txt' to deny web crawlers.
+    for_each = var.environment != "production" ? [1] : []
```
DFurnes, line 181: I just remembered to add this in ff77c81! 🙈 This keeps our same "deny all" `robots.txt` in development and QA to prevent those environments from showing up in Google search results. 

Terraform doesn't have a simple "conditional" syntax for blocks so, [like we do with the `count` attribute for resources](https://github.com/DoSomething/infrastructure/blob/13f1ec7367067a89746f605e75704aaa95eb397d/components/s3_bucket/main.tf#L113), we use a zero/one item "for loop" to make this conditional on `var.environment`.

dosomething-dev/main.tf
katiecrane: Oh smart prefixing there!

dosomething-dev/fastly-frontend/deny-robots.txt
```diff
@@ -1,14 +0,0 @@
...
-
-# Allow Facebook & Twitter crawlers for testing.
-User-agent: Facebot
-Disallow:
-
-User-agent: facebookexternalhit
-Disallow:
-
-User-agent: Twitterbot
-Disallow:
```
katiecrane, line 14: Generally I was seeing one file moved into a common folder and the two other copies deleted - for this one I only see two copies deleted, was that intentional?

katiecrane: Seeing same pattern with the takeovers, just two files removed

DFurnes: Oh I totally goofed (on including that file & properly testing the conditional)! Fixed in 7e02dfa.

dosomething-qa/fastly-frontend/takeover_config.vcl
```diff
@@ -1,8 +0,0 @@
-declare local var.takeover_start STRING;
-declare local var.takeover_end STRING;
-
-# This determines when a timed takeover is applied for this
-# environment (see `shared/takeover_recv.vcl`). Times must
-# be provided in 24-hour EST time.
-set var.takeover_start = "2018-11-07 00:00:00";
-set var.takeover_end = "2018-11-07 14:30:00";
```
katiecrane, line 8: What's up with these takeover files that were removed, are they just no longer needed?

DFurnes: Yep! They were for an old takeover we ran for the 2018 midterm elections.

I'd removed the config that referenced these in [`4fc0a59`](https://github.com/DoSomething/infrastructure/pull/270/files/4fc0a59fec55b0a62df45658bc6b7fd7dd3dda7e#diff-78e34551875d9e35121e22997ec0db55) but forgot to remove the files themselves.

katiecrane (approved):
good from me!

PR #274, opened 2020-09-21
Add standard headers to frontend & backend services.

DFurnes:
### What's this PR do?

This pull request adds standard security headers to our Fastly properties - ensuring that any application behind our CDN gets these applied! I've also separated out the variables where we set request & response headers (since unlike geolocation headers, we don't need these to be applied in both places).

You can see the changes made to the list of headers in our frontend & backend service components. These will then get applied consistently to all six properties when we merge & apply this.

🔒

### Relevant tickets

References [Pivotal #173719273](https://www.pivotaltracker.com/story/show/173719273).

### Checklist

### DIFF
--- a/components/fastly_backends/main.tf
+++ b/components/fastly_backends/main.tf
@@ -14,10 +14,38 @@ variable "papertrail_destination" {
 }
 
 locals {
-  headers = {
+  request_headers = {
+    # Provide geolocation information to web servers:
     "X-Fastly-Country-Code" = "client.geo.country_code",
     "X-Fastly-Region-Code"  = "client.geo.region",
     "X-Fastly-Postal-Code"  = "client.geo.postal_code",
+  }
+
+  response_headers = {
+    # Expose geolocation information in debug headers:
+    "X-Fastly-Country-Code" = "client.geo.country_code",
+    "X-Fastly-Region-Code"  = "client.geo.region",
+    "X-Fastly-Postal-Code"  = "client.geo.postal_code",
+
+    # Ensures that we don't leak sensitive items from URLs in Referrer headers: we will
+    # send full URL if same-origin and HTTPS, origin-only (e.g. 'www.dosomething.org')
+    # when crossing domains, and nothing at all when going to HTTP.
+    "Referrer-Policy" = "\"strict-origin-when-cross-origin\""
+
+    # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/174911098)
+    "Strict-Transport-Security" = "\"max-age=300; includeSubDomains\"" # 5 minutes.
+
+    # Prevent sniffing content types:
+    "X-Content-Type-Options" = "\"nosniff\""
+
+    # Restrict this application from being embedded as an iframe on
+    # third-party domains, to prevent click-jacking attacks:
+    "X-Frame-Options" = "\"SAMEORIGIN\""
+
+    # Enable XSS filtering in some browsers. This should be
+    # replaced by a strong Content Security Policy in the future.
+    # TODO: https://www.pivotaltracker.com/story/show/174911196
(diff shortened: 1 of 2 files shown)


components/fastly_backends/main.tf
```diff
@@ -14,11 +14,39 @@ variable "papertrail_destination" {
...
+  response_headers = {
+    # Expose geolocation information in debug headers:
+    "X-Fastly-Country-Code" = "client.geo.country_code",
+    "X-Fastly-Region-Code"  = "client.geo.region",
+    "X-Fastly-Postal-Code"  = "client.geo.postal_code",
+
+    # Ensures that we don't leak sensitive items from URLs in Referrer headers: we will
+    # send full URL if same-origin and HTTPS, origin-only (e.g. 'www.dosomething.org')
+    # when crossing domains, and nothing at all when going to HTTP.
+    "Referrer-Policy" = "\"strict-origin-when-cross-origin\""
```
DFurnes, line 33: To explain the funky quotes: by default Fastly interprets these values as variable references (like `client.geo.country_code` above). To set a literal string, we provide a quoted value (which, since strings in Terraform are themselves quoted, needs an escaped `\"` set of quotes inside the string).

components/fastly_backends/main.tf
```diff
@@ -14,11 +14,39 @@ variable "papertrail_destination" {
...
+
+    # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/174911098)
+    "Strict-Transport-Security" = "\"max-age=300; includeSubDomains\"" # 5 minutes.
```
DFurnes, line 36: Since this can't easily be undone once someone receives this header, we'll want to test this with a short TTL for a while, and then increase until we reach 2 years long following [these best practices](https://hstspreload.org/#deployment-recommendations). (And we can consider submitting our domain for HSTS preload then, too!)

components/fastly_frontend/main.tf
```diff
@@ -18,11 +18,39 @@ variable "papertrail_destination" {
...
+    "X-Content-Type-Options" = "\"nosniff\""
+
+    # Restrict this application from being embedded as an iframe on
+    # third-party domains, to prevent click-jacking attacks:
+    "X-Frame-Options" = "\"SAMEORIGIN\""
+
+    # Enable XSS filtering in some browsers. This should be
+    # replaced by a strong Content Security Policy in the future.
+    # TODO: https://www.pivotaltracker.com/story/show/174911196
+    "X-XSS-Protection" = "\"1\""
```
DFurnes, line 52: These are all identical to what's set in the `fastly_backend` module above! I think it's reasonable to keep these not-so-DRY, since we may eventually have differing security needs for backends and frontends.

components/fastly_backends/main.tf
katiecrane: Does this mean the header is only like active for 5 minutes? I'm a little confused about this

DFurnes: It's sort-of like a cache header – this header, as written here, will be sent on all responses from www.dosomething.org. When a user's browser sees it on a response, it'll make a note that it's _only_ allowed to visit our site via HTTPS for the next five minutes. (So if someone tries to visit `http://` directly or omits a protocol, it'll force that to `https://`).

After five minutes, it'll revert to default behavior (defaulting to `http://`) until it sees this header on another response & once again forces HTTPS traffic for the next five minutes. (And hence why we'll want to bump this up after testing.)

katiecrane: Ah got it! So if something does end up not working, it'll only be borked for five minutes?

DFurnes: Correct!

katiecrane (approved):
all my questions have been answered so good from me!

PR #275, opened 2020-09-24
Allow embedding page at end of voter registration flow.

DFurnes:
### What's this PR do?

This pull request removes the `X-Frame-Options` header from the `/us/about/voter-registration-thank-you` page embedded on the Rock The Vote "Thank You" page. This should allow this page to be embedded on this (and any other) third-party domains, while keeping protection enabled on other URLs.

👀

🇺🇸 

### Relevant tickets

References [Pivotal #174971084](https://www.pivotaltracker.com/story/show/174971084).

### Checklist

### DIFF
--- a/components/fastly_frontend/main.tf
+++ b/components/fastly_frontend/main.tf
@@ -134,6 +134,24 @@ resource "fastly_service_v1" "frontend" {
     }
   }
 
+  # Override: Remove 'X-Frame-Options' on URL(s) that we do
+  # want to allow embedding from external domains:
+  condition {
+    type      = "RESPONSE"
+    name      = "allow-embed"
+    statement = "req.url ~ \"^/us/about/voter-registration-thank-you\""
+  }
+
+  header {
+    name               = "Allow Embed on Path(s)"
+    type               = "response"
+    action             = "delete"
+    destination        = "http.X-Frame-Options"
+    response_condition = "allow-embed"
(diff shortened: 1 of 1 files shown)


components/fastly_frontend/main.tf
```diff
@@ -134,6 +134,24 @@ resource "fastly_service_v1" "frontend" {
...
+    name      = "allow-embed"
+    statement = "req.url ~ \"^/us/about/voter-registration-thank-you\""
+  }
+
+  header {
+    name               = "Allow Embed on Path(s)"
+    type               = "response"
+    action             = "delete"
+    destination        = "http.X-Frame-Options"
+    response_condition = "allow-embed"
```
katiecrane, line 150: This `response_condition` is what ties the two together? Not super familiar with the syntax of this

DFurnes: Yep, exactly! By default, changes like this are applied to all requests. The `condition` block allows us to segment something on a subset of requests/responses, and then we can attach that to other behavior.

We also use this to [route traffic to different backends](https://github.com/DoSomething/infrastructure/blob/daec9a076a6f4f5370f0d4d2adc182ac753c50f2/components/fastly_backends/main.tf#L79-L91) based on `req.host` in our backend property.

katiecrane: Ah those examples are helpful, thanks!

katiecrane (approved):
looks good to me!

PR #277, opened 2020-09-24
Removing backup window to allow RDS to manage this

blisteringherb:
### What's this PR do?

This pull request removes the backup window variable to allow RDS to manage this. The `apply` errored out because the window overlaps the maintenance window. At the moment we don't need to have fine-grained control over that, and we can just let the existing window stand.

...

https://app.terraform.io/app/dosomething/workspaces/quasar/runs/run-Cvt1FajKA1u7BvNn

### Relevant tickets

### Checklist

### DIFF
--- a/components/postgresql_warehouse/main.tf
+++ b/components/postgresql_warehouse/main.tf
@@ -178,7 +178,6 @@ resource "aws_db_instance" "quasar" {
   engine_version                  = var.engine_version
   allow_major_version_upgrade     = true
   backup_retention_period         = local.backup_retention # See above!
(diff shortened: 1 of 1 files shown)


DFurnes (approved):

blisteringherb:
@DFurnes it looks like the check for dosomething-prod is taking an exceptionally long time. Have you run into something like this before? The voting-app check failed the last time on this PR, and I tried to restart it but got an error.

DFurnes:
I haven't run into that issue before. It looks like a request 503'd for that particular plan.

I tried pushing an empty commit to re-run these plans & see if that fixes things.

PR #278, opened 2020-10-05
Adding quasar s3 role and policy

blisteringherb:
### What's this PR do?

This PR creates the IAM role and policy for the RDS Quasar snapshots to be stored in S3, and a new bucket to store the archived snapshots.

...

...

### Relevant tickets

### Checklist

### DIFF
--- a/components/quasar_s3_export_role/main.tf
+++ b/components/quasar_s3_export_role/main.tf
@@ -52,7 +52,7 @@ data "aws_iam_policy_document" "quasar_s3_export_role" {
 
 resource "aws_iam_policy" "policy" {
   name   = "ExportPolicy"
-  policy = templatefile("${path.module}/policy.json.tpl", { bucket_arn = aws_s3_bucket.dosomething_quasar_archive.arn })
+  policy = templatefile("${path.module}/iam-policy.json.tpl", { bucket_arn = aws_s3_bucket.dosomething_quasar_archive.arn })
 }
 
 resource "aws_iam_role" "role" {
--- a/components/quasar_s3_export_role/main.tf
+++ b/components/quasar_s3_export_role/main.tf
@@ -22,7 +22,7 @@ data "aws_iam_policy_document" "quasar_s3_export_role" {
 
 resource "aws_iam_policy" "policy" {
   name   = "ExportPolicy"
-  policy = templatefile("${path.module}/iam-policy.json.tpl", { bucket_arn = aws_s3_bucket.dosomething_quasar_archive.arn })
+  policy = templatefile("${path.module}/iam-policy.json.tpl", { bucket_arn = var.arn })
 }
 
 resource "aws_iam_role" "role" {
--- a/components/quasar_s3_export_role/main.tf
+++ b/components/quasar_s3_export_role/main.tf
@@ -1,0 +1,28 @@
+data "aws_iam_policy_document" "quasar_s3_export_role" {
+  statement {
+    effect = "Allow"
+    actions = ["sts:AssumeRole"]
+    principals {
+      type        = "Service"
+      identifiers = ["export.rds.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_policy" "policy" {
+  name = "ExportPolicy"
+  policy = templatefile("${path.module}/iam-policy.json.tpl", {
+    bucket_arn = var.bucket.arn
+  })
+}
+
+resource "aws_iam_role" "role" {
+  name               = "rds-s3-export-role"
+  description        = "Role for exporting Quasar RDS snapshots to S3."
+  assume_role_policy = data.aws_iam_policy_document.quasar_s3_export_role.json
+}
+
+resource "aws_iam_role_policy_attachment" "attachment" {
+  role       = "${aws_iam_role.role.name}"
+  policy_arn = "${aws_iam_policy.policy.arn}"
+}
--- a/components/quasar_s3_export_role/main.tf
+++ b/components/quasar_s3_export_role/main.tf
@@ -1,3 +1,44 @@
+resource "aws_s3_bucket" "dosomething_quasar_archive" {
+  bucket = "dosomething-quasar-archive"
+  acl    = "private"
+
+  lifecycle_rule {
+    id      = "archive"
+    enabled = true
+
+    prefix = "archive/"
+
+    tags = {
+      "rule" = "archive"
+    }
+
+    transition {
+      days          = 30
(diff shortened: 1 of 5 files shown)


components/quasar_s3_export_role/main.tf
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    transition {
+      days          = 30
+      storage_class = "GLACIER"
+    }
+
+    expiration {
+      days = 60
+    }
```
DFurnes, line 27: Does this mean that these archived objects will be deleted after 60 days?

components/quasar_s3_export_role/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+resource "aws_s3_bucket_public_access_block" "private_policy" {
+  bucket = aws_s3_bucket.dosomething_quasar_archive.id
+
+  # We don't want to risk a bug causing individual
+  # log files to be marked with "public" ACLs.
+  block_public_acls  = true
+  ignore_public_acls = true
```
DFurnes, line 38: I think we can add `block_public_policy = true` here too (which prevents bucket policies from accidentally "opening up" this bucket's contents if we'd made a mistake there).

components/quasar_s3_export_role/main.tf
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+  }
+}
+
+resource "aws_s3_bucket_public_access_block" "private_policy" {
+  bucket = aws_s3_bucket.dosomething_quasar_archive.id
+
+  # We don't want to risk a bug causing individual
+  # log files to be marked with "public" ACLs.
```
DFurnes, line 36:
```suggestion
  # archived files to be marked with "public" ACLs.
```

quasar/main.tf
```diff
@@ -118,3 +118,8 @@ module "storage" {
   versioning = true
   private    = true
 }
+
+module "quasar_s3_export_role" {
```
DFurnes, line 122: Based on the name of this module, I wonder if it should _just_ have the role in it?

We could potentially use our `s3_bucket` module to create the Glacier bucket and set up archive rules:

```hcl
module "quasar_archive" {
  source = "../components/s3_bucket"

  application = "dosomething-quasar"
  name        = "dosomething-quasar-archive"
  environment = "production"
  stack       = "data"

  versioning = true
  archived   = true
  private    = true
}

module "rds_export_role" {
  source = "../components/quasar_s3_export_role"
  
  arn = module.quasar_archive.arn
}
```

components/quasar_s3_export_role/main.tf
blisteringherb: Yes, they'll expire after 60 days. Would you prefer a longer time period?

DFurnes: I think I'd misunderstood this task, since I'd assumed the reason we were putting these into Glacier was for long-term storage. Is this something you're planning on using for RDS's daily snapshots?

blisteringherb: I was thinking that we'd push the snapshots into s3 daily and then push them into Glacier after a set period, and eventually expire them since we already have the backup in Glacier if we really need to pull it out.

DFurnes: I believe the Glacier lifecycle rule "moves" those objects into Glacier (as their storage class in S3), but the objects themselves are still represented as S3 items. So in this case, these rules would be moving the objects into Glacier and then deleting them (from both S3 & Glacier) 30 days later.

DFurnes: If we want to store these indefinitely in Glacier, we don't need an `expiration` rule here. If we _do_ want to expire these so these backups don't pile up over time, we should set an `expiration` rule of [at least 90 days](https://docs.aws.amazon.com/AmazonS3/latest/dev/lifecycle-transition-general-considerations.html#before-deciding-to-archive-objects) or else it looks like we'll be charged an "early deletion fee".

components/quasar_s3_export_role/main.tf
```diff
@@ -0,0 +1,67 @@
+resource "aws_s3_bucket" "dosomething_quasar_archive" {
+  bucket = "dosomething-quasar-archive"
```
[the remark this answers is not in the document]
blisteringherb: On some further reflection, I think it does make sense to put these in a separate bucket since these files will be parquet files and may end up being the beginning of our data lake. If we want to query the snapshots we'll probably use something like Athena, so it would make sense to separate the snapshot files from other types of backup files.

blisteringherb:
@DFurnes I'm getting this error:
```
Error: Reference to undeclared output value

  on main.tf line 138, in module "rds_export_role":
 138:   arn = module.quasar_archive.role_arn

An output value with the name "role_arn" has not been declared in
module.quasar_archive.
```

Here's the declaration in `quasar/main.tf` for creating the s3 bucket. The output in the `s3_bucket` seems to indicate that there's a `role_arn` variable available, but it looks like I'm referencing it incorrectly. Do you have any insight?

```
module "quasar_archive" {
  source = "../components/s3_bucket"

  application = "dosomething-quasar"
  name        = "dosomething-quasar-archive"
  environment = "production"
  stack       = "data"

  versioning = true
  archived   = true
  private    = true
}

module "rds_export_role" {
  source = "../components/quasar_s3_export_role"

  arn = module.quasar_archive.role_arn
}
```

DFurnes:
You're exporting `role_arn` from the `rds_export_role` module (defined [here](https://github.com/DoSomething/infrastructure/blob/1867cf44ac1f3dc3630c9c2b8337c5dc02b3f7fe/components/quasar_s3_export_role/outputs.tf)), so you can't use that as an export on the `s3_bucket` module (which defines it's own exports [here](https://github.com/DoSomething/infrastructure/blob/1867cf44ac1f3dc3630c9c2b8337c5dc02b3f7fe/components/s3_bucket/main.tf#L189-L212)). Each module has its own specific "inputs" and "outputs".

To provide the S3 bucket's ARN to this role module, you could do this:

```hcl
module "rds_export_role" {
  source = "../components/quasar_s3_export_role"

  arn = module.quasar_archive.bucket.arn
}
```

components/quasar_s3_export_role/main.tf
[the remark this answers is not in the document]
DFurnes: Ah, that makes great sense to me!

DFurnes (approved):
Looks great!

PR #279, opened 2020-10-07
Configure Gambit URL & credentials for this app.

DFurnes:
### What's this PR do?

This pull request configures the required `GAMBIT_URL`, `GAMBIT_USERNAME`, and `GAMBIT_PASSWORD` environment variables for the queue job introduced in DoSomething/rogue#1128. It also sets the `DS_ENABLE_BLINK`, `DS_ENABLE_CUSTOMER_IO`, and `DS_ENABLE_GAMBIT_RELAY` feature flags based on environment.

Does this set the right variables? Do the plans succeed?

This application is taking some responsibility (to talk directly to Gambit) away from Blink. We want to make these changes here in Terraform so that they're easy to replicate if we need to rebuild any of these environments from scratch, and so that we don't accidentally apply the wrong settings if making these changes by hand.

### Relevant tickets

References [Pivotal #174827024](https://www.pivotaltracker.com/story/show/174827024).

### Checklist

### DIFF
--- a/applications/rogue/main.tf
+++ b/applications/rogue/main.tf
@@ -54,6 +54,12 @@ variable "graphql_url" {
 
 variable "blink_url" {
   description = "The Blink URL for this environment."
+  default     = null
+}
+
+variable "gambit_url" {
+  description = "The Gambit URL for this environment."
+  default     = null
 }
 
 variable "papertrail_destination" {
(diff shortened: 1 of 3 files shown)


applications/rogue/main.tf
```diff
@@ -81,20 +87,41 @@ resource "random_id" "app_key" {
 }
 
 locals {
+  # Gambit only has QA & Production environments.
+  gambit_env = var.environment == "development" ? "qa" : var.environment
```
DFurnes, line 91: Context: As far as I know, I can't cleanly make the data resources (to read the username/password from SSM) optional, and so it's easiest to just plop the QA credentials onto the dev environment & disable via feature flag.

katiecrane: Is the feature flag happening in this file? I would have thought `DS_ENABLE_GAMBIT_RELAY` would be false for dev?

katiecrane: OH I see, you only added the Gambit url to prod and QA yeah?

DFurnes: Yep!

katiecrane (approved):
looks good!

PR #281, opened 2020-10-15
Increase 'Strict-Transport-Security' duration.

DFurnes:
### What's this PR do?

This pull request increases our `Strict-Transport-Security` duration to 1 week. This means that once a user visits our website (and their browser sees this header), they'll _only_ be able to load `*.dosomething.org` sites over HTTPS. (For [the past few weeks](https://github.com/DoSomething/infrastructure/pull/274/files/b34a1bd79e7d3b5b8930408434d07158a84c68e4#r492348243), HTTPS would only be forced for the following 5 minutes after a user visited the site.)

👀

We're slowly ramping this up over time to make sure that we don't accidentally cause issues we can't fix (since the setting is stored on user's local machines and therefore there's no way to "undo" this aside from waiting the duration specified).

### Relevant tickets

References [Pivotal #174911098](https://www.pivotaltracker.com/story/show/174911098).

### Checklist

### DIFF
--- a/components/fastly_backends/main.tf
+++ b/components/fastly_backends/main.tf
@@ -32,7 +32,7 @@ locals {
     # when crossing domains, and nothing at all when going to HTTP.
     "Referrer-Policy" = "\"strict-origin-when-cross-origin\""
 
-    # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/174911098)
+    # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/175288875)
     "Strict-Transport-Security" = "\"max-age=604800; includeSubDomains\"" # 1 week.
 
     # Prevent sniffing content types:
--- a/components/fastly_backends/main.tf
+++ b/components/fastly_backends/main.tf
@@ -33,7 +33,7 @@ locals {
     "Referrer-Policy" = "\"strict-origin-when-cross-origin\""
 
     # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/174911098)
-    "Strict-Transport-Security" = "\"max-age=300; includeSubDomains\"" # 5 minutes.
+    "Strict-Transport-Security" = "\"max-age=604800; includeSubDomains\"" # 1 week.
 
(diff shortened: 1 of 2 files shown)


components/fastly_backends/main.tf
```diff
@@ -32,8 +32,8 @@ locals {
     # when crossing domains, and nothing at all when going to HTTP.
     "Referrer-Policy" = "\"strict-origin-when-cross-origin\""
 
-    # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/174911098)
-    "Strict-Transport-Security" = "\"max-age=300; includeSubDomains\"" # 5 minutes.
+    # Enforce HTTPS for all traffic (TODO: https://www.pivotaltracker.com/story/show/175288875)
+    "Strict-Transport-Security" = "\"max-age=604800; includeSubDomains\"" # 1 week.
```
katiecrane, line 36: Is the thought on this bump that if we were gonna see issues, we would have seen them with 5 minutes? So it's okay to jump straight to a week? What kind of errors would we have seen if there were issues?

DFurnes: Yeah, I went back and forth on whether to try something like a day or 3 days... this is just following Google's [deployment recommendations](https://hstspreload.org/#deployment-recommendations) from [hstspreload.org](https://hstspreload.org/).

If this didn't work, we'd be seeing Zendesk or staff reports of errors (like a "cannot connect" Chrome tab or something) where users couldn't load a `*.dosomething.org` subdomain that didn't support HTTPS. I don't know any off the top of my head, so I _think_ this unlikely though!

katiecrane: @DFurnes ah okay so 5 minutes to 1 week is standard practice then! I'm okay with that, just was hesitant since we can't undo it

katiecrane (approved):
looks good!

components/fastly_backends/main.tf
DFurnes: Yeah, same! I think this should hopefully be pretty low-risk, and we'll wait a while before we bump it to a month. 🤞

PR #282, opened 2020-10-23
Add DS_ENABLE_CUSTOMER_IO flag to Northstar.

DFurnes:
### What's this PR do?

This pull request simply adds the `DS_ENABLE_CUSTOMER_IO` feature flag to Northstar, so it can replace the old (and no longer accurately named) `DS_ENABLE_BLINK` feature flag [like we've done in Rogue](https://github.com/DoSomething/infrastructure/blob/8864abc714907d9ba735429c159b290016270e98/applications/rogue/main.tf#L109).

👀

We had previously not been handling this feature flag in Terraform, but it's a good time to start!

### Relevant tickets

N/A

### Checklist

### DIFF
--- a/applications/northstar/main.tf
+++ b/applications/northstar/main.tf
@@ -91,6 +91,7 @@ locals {
   feature_config_vars = {
     DS_ENABLE_PASSWORD_GRANT = false
     DS_ENABLE_RATE_LIMITING  = true
+    DS_ENABLE_CUSTOMER_IO    = var.environment != "development"
   }
 
   # This application is part of our backend stack.


applications/northstar/main.tf
```diff
@@ -91,6 +91,7 @@ locals {
   feature_config_vars = {
     DS_ENABLE_PASSWORD_GRANT = false
     DS_ENABLE_RATE_LIMITING  = true
+    DS_ENABLE_CUSTOMER_IO    = var.environment != "development"
```
katiecrane, line 94: Can you remind me of how this works? This will just set it up in Heroku for us, but we can still edit it via Heroku?

DFurnes: We will still be able to edit it via Heroku, but we probably won't want to outside of an emergency. Instead, we'd want to let Terraform manage these environment variables, both to set them initially & to "reset" them if they're edited by hand (so we can avoid [configuration drift](https://www.hashicorp.com/resources/how-can-i-prevent-configuration-drift)).

katiecrane: Awesome!

katiecrane (approved):
looks good!

PR #284, opened 2020-11-10
Changing production backup retention to 7 days in prod

blisteringherb:
### What's this PR do?

This pull request changes the backup retention period for production snapshots from 14 days to 7 days.

...

...

### Relevant tickets

### Checklist

### DIFF
--- a/components/postgresql_warehouse/main.tf
+++ b/components/postgresql_warehouse/main.tf
@@ -53,9 +53,9 @@ variable "vpc_security_group_ids" {
 locals {
   stack = "data"
 
-  # We keep backups for 14 days in producion, and otherwise the minimum
+  # We keep backups for 7 days in producion, and otherwise the minimum
   # amount in order to satisfy functionality.
-  backup_retention = var.environment == "production" ? 14 : 1
+  backup_retention = var.environment == "production" ? 7 : 1
 }
 
 resource "aws_db_parameter_group" "pg11" {


DFurnes (approved):

rapala61:
Hi @DFurnes, I'm not super familiar with Terraform, but have a question about this change. Do you remember if the change was actually deployed to Production? I see that our Quasar RDS continues to save 14 days vs the 7 days configured here. I wonder if there is another thing we need to do so it takes effect. Thanks in advance!

DFurnes:
@rapala61 It looks like it wasn't! I've sent you over a Terraform Cloud invite, and you can see the pending plan for this workspace [here](https://app.terraform.io/app/dosomething/workspaces/quasar/runs/run-7MKxgfy8pWQcDUWj). If everything looks good, just hit the "Confirm & Apply" button to apply that change in production.

rapala61:
@DFurnes I'm looking at the [changes](https://app.terraform.io/app/dosomething/workspaces/quasar/runs/run-7MKxgfy8pWQcDUWj) and it looks good as far as changing the backup retention from 14 to 7 days. However, it includes an additional change that looks a bit suspicious to me:
- Prod engine version: `engine_version                        = "11.8" -> "11.4"`
- QA engine version: `engine_version                        = "11.8" -> "11.6"`

This seems like a versioning pinning issue. I certainly don't intend to downgrade the engine. What do you suggest in this case?

DFurnes:
Mmm, sounds like configuration drift! If that was an intentional change, let's reflect it in our configuration [here](https://github.com/DoSomething/infrastructure/blob/d0b7bc34f08a67333814edc9f5b91204383be81c/quasar/main.tf#L66) and [here](https://github.com/DoSomething/infrastructure/blob/d0b7bc34f08a67333814edc9f5b91204383be81c/quasar/main.tf#L87).

PR #286, opened 2021-02-08
Use same S3 bucket for Northstar & Rogue.

DFurnes:
### What's this PR do?

This pull request moves Rogue's storage buckets into Northstar's application module (as we're combining these two codebases). I've already moved the files in each of Northstar's buckets into it's Rogue equivalent.

👀

This allows Northstar to read/write uploads while continuing to access it's own files!

### Relevant tickets

References [Pivotal #176851342](https://www.pivotaltracker.com/story/show/176851342).

### Checklist

### DIFF
--- a/applications/northstar/main.tf
+++ b/applications/northstar/main.tf
@@ -35,6 +35,15 @@ variable "name" {
   description = "The application name."
 }
 
+variable "storage_name" {
+  description = "The storage bucket name."
+}
+
+variable "backup_storage_bucket" {
+  description = "Optionally, the bucket to replicate storage backups to."
+  default     = null
+}
+
 variable "papertrail_destination" {
   description = "The Papertrail log destination for this application."
 }
@@ -160,13 +169,14 @@ module "storage" {
   source = "../../components/s3_bucket"
 
   application = var.name
-  name        = var.name
+  name        = var.storage_name
   environment = var.environment
   stack       = local.stack
+  user        = module.iam_user.name
 
-  user       = module.iam_user.name
-  private    = true
-  versioning = true
+  private            = true
+  replication_target = var.backup_storage_bucket
+  versioning         = true
 }
 
 output "name" {
--- a/applications/northstar/main.tf
+++ b/applications/northstar/main.tf
@@ -176,7 +176,7 @@ module "storage" {
 
   private            = true
   replication_target = var.backup_storage_bucket
-  versioning         = true
+  versioning         = var.environment == "production"
 }
 
 output "name" {
--- a/applications/rogue/main.tf
+++ b/applications/rogue/main.tf
@@ -66,11 +66,6 @@ variable "papertrail_destination" {
   description = "The Papertrail log destination for this application."
 }
 
-variable "backup_storage_bucket" {
-  description = "Optionally, the bucket to replicate storage backups to."
-  default     = null
-}
-
 variable "deprecated" {
   description = "Deprecate this app, removing database users & allowing deletion."
   default     = false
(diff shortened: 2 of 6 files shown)


applications/northstar/main.tf
```diff
@@ -35,6 +35,15 @@ variable "name" {
   description = "The application name."
 }
 
+variable "storage_name" {
+  description = "The storage bucket name."
+}
```
DFurnes, line 40: This is necessary since the bucket's name (e.g. `dosomething-rogue`) will no longer match the application name (e.g. `dosomething-northstar`). We can consider cleaning this up later, but it'd be a timely migration since you can't rename existing buckets.

applications/northstar/main.tf
```diff
@@ -35,6 +35,15 @@ variable "name" {
...
+
+variable "backup_storage_bucket" {
+  description = "Optionally, the bucket to replicate storage backups to."
+  default     = null
+}
```
DFurnes, line 45: This is a variable we used to provide to Rogue to allow us to configure our cross-region backup bucket in production. Since this bucket is now living under Northstar, we provide this variable here instead!

applications/rogue/main.tf
```diff
@@ -108,6 +103,14 @@ locals {
...
     DS_ENABLE_CUSTOMER_IO  = var.environment != "development"
     DS_ENABLE_GAMBIT_RELAY = var.gambit_url != null
+
+    # S3 settings:
+    STORAGE_DRIVER    = "s3"
+    FILESYSTEM_DRIVER = "s3"
+    S3_BUCKET         = var.name
+    AWS_S3_BUCKET     = var.name
+    S3_REGION         = "us-east-1"
+    AWS_S3_REGION     = "us-east-1"
```
DFurnes, line 113: Since we're removing the S3 bucket module, which used to provide these, I've hard-coded them under `extra_config_vars` for now to make everything continues to function exactly the same way.

applications/rogue/main.tf
```diff
@@ -182,17 +184,16 @@ module "queue" {
...
+}
 
+resource "aws_iam_user_policy" "s3_policy" {
+  name = "${var.name}-s3"
   user = module.iam_user.name
-
-  replication_target = var.backup_storage_bucket
+  policy = templatefile("${path.module}/s3-iam-policy.json.tpl", {
+    bucket_arn = data.aws_s3_bucket.bucket.arn
+  })
```
DFurnes, line 196: I replaced the [`module.s3_bucket`](https://github.com/DoSomething/infrastructure/tree/main/components/s3_bucket) module here with a [`s3_bucket` data resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/s3_bucket) (which _reads_ from an existing bucket by name, rather than creating/modifying it).

This allows us to continue to reference the bucket that we've moved into Northstar's application module. I then created a `aws_iam_user_policy.s3_policy` that allows Rogue's IAM user to continue to access this bucket (since the relocated module will now default to granting access to Northstar).

DFurnes: Using a data resource for something that's defined elsewhere is probably a bad practice (as opposed to defining the bucket elsewhere and passing it in as a variable to each app that shares it), but I think this the cleanest way to handle this during this "transition period".

applications/rogue/s3-iam-policy.json.tpl
```diff
@@ -0,0 +1,16 @@
...
+            "Action": [
+                "s3:*"
+            ],
+            "Resource": [
+                "${bucket_arn}/*",
+                "${bucket_arn}"
+            ]
+        }
+    ]
+}
```
DFurnes, line 16: This is a simple copy-paste from the [`s3_bucket`](https://github.com/DoSomething/infrastructure/blob/main/components/s3_bucket/iam-policy.json.tpl) module (because each module only has access to it's own folder scope).

katiecrane (approved):
I'm not super familiar with this, but your comments are very helpful! Is anything actually changing with the Rogue buckets or we just set them up from inside Northstar now?

DFurnes:
Yeah, there shouldn't 🤞  be any changes from Rogue's perspective! We're just configuring them from "within" Northstar's module & now reading from that bucket in Northstar for OAuth keys (so we don't have to read & write to two different S3 buckets in Northstar).

katiecrane:
@DFurnes perfect! Sounds like a good way to do it!

PR #290, opened 2021-04-15
Add instapage redirects to phoenix

lindsayrmaher:
### What's this PR do?

This pull request adds domains from instapage to redirect to their corresponding page on phoenix!

👀 

We're moving off of instapage and want to make sure anyone who tries to use those links is sent to our site instead!

### Relevant tickets

References [Pivotal #177518771](https://www.pivotaltracker.com/story/show/177518771).

### Checklist

### DIFF
--- a/redirects/redirects_table.vcl
+++ b/redirects/redirects_table.vcl
@@ -60,4 +60,9 @@ table redirects {
   "www.fourleggedfinishers.com": "https://www.dosomething.org/us/campaigns/four-legged-finishers#",
   "www.fourleggedfinishers.org": "https://www.dosomething.org/us/campaigns/four-legged-finishers#",
   "www.catsgonegood.com": "https://www.celebsgonegood.com",
+
+  # instapage:
+  "lets.dosomething.org": "https://www.dosomething.org/us/articles#",
+  "join.dosomething.org": "https://www.dosomething.org/us/about#",
+  "vote.dosomething.org": "https://www.dosomething.org/us/vote#",
 }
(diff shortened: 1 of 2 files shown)


redirects/redirects_table.vcl
```diff
@@ -60,4 +60,9 @@ table redirects {
   "www.fourleggedfinishers.com": "https://www.dosomething.org/us/campaigns/four-legged-finishers#",
   "www.fourleggedfinishers.org": "https://www.dosomething.org/us/campaigns/four-legged-finishers#",
   "www.catsgonegood.com": "https://www.celebsgonegood.com",
+
+  # instapage:
+  "lets.dosomething.org": "https://www.dosomething.org/us/articles#",
```
katiecrane, line 65: What are the `#`s here for? I see some of our redirects have that and some don't and I'm not familiar with that syntax!

lindsayrmaher: @katiecrane  it's a hacky lil workaround for making sure different paths that have the same origin location like `vote.dosomething.org` and `vote.dosomething.org/xyz` end up redirecting to the same place because the # breaks the path. We definitely need it for the VR pages but figured it couldn't hurt to add in on the other redirects as well in case someone tries to go to a different path. cc @DFurnes

katiecrane: Oh that's cool! Nice that there is an easy way to get that behavior and agreed that it couldn't hurt on the others too.

DFurnes (approved):

PR #291, opened 2021-04-15
Ghost Inspector Heroku Deploy Hook

mendelB:
### What's this PR do?

This pull request adds a new "Ghost Inspector Webhook" component which configures a Heroku ([HTTP Post](https://devcenter.heroku.com/articles/deploy-hooks#http-post-hook)) Deploy Hook addon to automatically kick off our [Ghost Inspector QA suite](https://app.ghostinspector.com/suites/56980f16812f6dd51bc172f2) (using the webhook address stored in an AWS SSM Parameter) following an application deployment.

It's configured on the Phoenix & Northstar Heroku applications.

👀 

This will help ensure that our team is on top of any test failures following a Pull Request being merged (and the app deploying to our QA environments).

Co-authored by @DFurnes 🎩 

### Relevant tickets

References [Pivotal #170523172](https://www.pivotaltracker.com/story/show/170523172).

### Checklist

### DIFF
--- a/components/ghost_inspector_webhook/main.tf
+++ b/components/ghost_inspector_webhook/main.tf
@@ -1,0 +1,14 @@
+data "aws_ssm_parameter" "ghost_inspector_webhook" {
+  count = var.environment == "qa" ? 1 : 0
+  name  = "/ghost-inspector/webhooks/qa"
+}
+
+resource "heroku_addon" "webhook" {
+  count = var.environment == "qa" ? 1 : 0
+  app   = var.name
+  plan  = "deployhooks:http"
+
+  config = {
+    url = data.aws_ssm_parameter.ghost_inspector_webhook[0].value
+  }
+}
--- a/applications/northstar/main.tf
+++ b/applications/northstar/main.tf
@@ -151,6 +151,12 @@ module "app" {
   with_newrelic          = coalesce(var.with_newrelic, var.environment == "production")
 }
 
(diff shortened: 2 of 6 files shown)


components/ghost_inspector_webhook/main.tf
```diff
@@ -0,0 +1,14 @@
+data "aws_ssm_parameter" "ghost_inspector_webhook" {
+  count = var.environment == "qa" ? 1 : 0
```
katiecrane, line 2: I remember this `count` is a workaround way to do something but I need a refresher on what's happening here!

mendelB: My understanding is that this Terraform defines [`count`](https://www.terraform.io/docs/language/meta-arguments/count.html#basic-syntax) amount of resources. So it's our way of "turning it on" only for QA since on prod it would be `0`.

cc @DFurnes I could be completely misunderstanding!

katiecrane: Ah that makes sense in this context to me!

katiecrane (approved):
looks good to me!

components/ghost_inspector_webhook/main.tf
DFurnes: Yes, that's exactly it!

DFurnes (approved):
Looks great!

PR #295, opened 2021-05-17
Configure lifecycle rules to automatically clear out Chompy uploads.

DFurnes:
### What's this PR do?

This pull request configures an S3 [lifecycle rule](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html) to automatically delete files uploaded to the `temporary/` directory in Northstar's S3 bucket.

👀

This allows us to ensure that these files don't pile up over time! 📚

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/components/s3_bucket/main.tf
+++ b/components/s3_bucket/main.tf
@@ -27,6 +27,12 @@ variable "roles" {
   default     = []
 }
 
+variable "temporary_paths" {
+  description = "Automatically delete files in the following paths after 14 days."
+  type        = list(string)
+  default     = []
+}
+
 variable "versioning" {
   description = "Enable versioning on this bucket. See: https://goo.gl/idPRVV"
   default     = false
@@ -60,6 +66,26 @@ resource "aws_s3_bucket" "bucket" {
   versioning {
     # Versioning must be enabled if we have a replication target.
     enabled = var.versioning || var.replication_target != null
+  }
+
+  dynamic "lifecycle_rule" {
+    for_each = var.temporary_paths
+
+    content {
+      id      = md5(lifecycle_rule.value)
+      prefix  = lifecycle_rule.value
+      enabled = true
+
+      # Delete files in this path after two weeks:
+      expiration {
+        days = 14
+      }
+
+      # If versioned, delete non-current versions at this path after 3 months:
+      noncurrent_version_expiration {
+        days = 90
+      }
+    }
   }
 
   dynamic "lifecycle_rule" {
--- a/components/s3_bucket/main.tf
+++ b/components/s3_bucket/main.tf
@@ -72,7 +72,6 @@ resource "aws_s3_bucket" "bucket" {
     for_each = var.temporary_paths
 
     content {
-      id      = md5(lifecycle_rule.value)
       prefix  = lifecycle_rule.value
       enabled = true
 
@@ -93,7 +92,6 @@ resource "aws_s3_bucket" "bucket" {
     for_each = var.archived ? [1] : []
 
     content {
-      id      = random_id.lifecycle_rule_id[0].b64_std
       enabled = true
 
       # Since we can't replicate directly into Glacier, set a lifecycle
@@ -116,7 +114,6 @@ resource "aws_s3_bucket" "bucket" {
       role = aws_iam_role.replication[0].arn
 
       rules {
-        id     = random_id.replication_rules[0].b64_std
         status = "Enabled"
 
         destination {
@@ -132,12 +129,6 @@ resource "aws_s3_bucket" "bucket" {
     Environment = var.environment
     Stack       = var.stack
   }
-}
-
-# Glacier:
-resource "random_id" "lifecycle_rule_id" {
-  count       = var.archived == true ? 1 : 0
-  byte_length = 32
 }
 
 # IAM policy:
@@ -174,11 +165,6 @@ resource "aws_s3_bucket_public_access_block" "private_policy" {
 }
 
 # Replication rules:
-resource "random_id" "replication_rules" {
-  count       = var.replication_target != null ? 1 : 0
-  byte_length = 32
-}
-
 resource "aws_iam_role" "replication" {
   count = var.replication_target != null ? 1 : 0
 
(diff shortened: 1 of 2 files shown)


components/s3_bucket/main.tf
```diff
@@ -27,6 +27,12 @@ variable "roles" {
   default     = []
 }
 
+variable "temporary_paths" {
+  description = "Automatically delete files in the following paths after 14 days."
+  type        = list(string)
+  default     = []
```
DFurnes, line 33: This says that, by default, we don't have any temporary paths. If we set this variable, it has to be an array of strings.

components/s3_bucket/main.tf
```diff
@@ -62,6 +68,26 @@ resource "aws_s3_bucket" "bucket" {
     enabled = var.versioning || var.replication_target != null
   }
 
+  dynamic "lifecycle_rule" {
+    for_each = var.temporary_paths
+
+    content {
+      id      = md5(lifecycle_rule.value)
```
DFurnes, line 75: Hmm, I'd added this because (as far as I remembered) this argument was required in the past! When I went to link to that documentation, though, it says it's optional! I'm going to try omitting and see how it goes.

components/s3_bucket/main.tf
```diff
@@ -62,6 +68,26 @@ resource "aws_s3_bucket" "bucket" {
     enabled = var.versioning || var.replication_target != null
   }
 
+  dynamic "lifecycle_rule" {
```
DFurnes, line 71: Because of how this `aws_s3_bucket` resource was built by Hashicorp/AWS, we have to configure this lifecycle rule as a dynamic "property" on the `aws_s3_bucket` (rather than as it's own optional resource that we could point to the bucket it's being attached to, like we do with a lot of other things).

Terraform's dynamic blocks are described in more detail [here](https://www.terraform.io/docs/language/expressions/dynamic-blocks.html).

components/s3_bucket/main.tf
```diff
@@ -62,6 +68,26 @@ resource "aws_s3_bucket" "bucket" {
...
+
+      # Delete files in this path after two weeks:
+      expiration {
+        days = 14
+      }
+
+      # If versioned, delete non-current versions at this path after 3 months:
+      noncurrent_version_expiration {
+        days = 90
+      }
```
DFurnes, line 87: Since we have versioning enabled on this production bucket, we'd end up with a lot of "deleted" version markers inside this `temporary/` directory. I figured it made sense to clean these up 90 days after the file itself is deleted. (We'll continue to retain other old versions indefinitely for safe-keeping).

components/s3_bucket/main.tf
```diff
@@ -62,6 +68,26 @@ resource "aws_s3_bucket" "bucket" {
...
+      id      = md5(lifecycle_rule.value)
+      prefix  = lifecycle_rule.value
+      enabled = true
+
+      # Delete files in this path after two weeks:
+      expiration {
+        days = 14
+      }
+
+      # If versioned, delete non-current versions at this path after 3 months:
```
katiecrane, line 84: What does this mean?

components/s3_bucket/main.tf
```diff
@@ -62,6 +68,26 @@ resource "aws_s3_bucket" "bucket" {
     enabled = var.versioning || var.replication_target != null
   }
 
+  dynamic "lifecycle_rule" {
+    for_each = var.temporary_paths
+
+    content {
+      id      = md5(lifecycle_rule.value)
+      prefix  = lifecycle_rule.value
```
katiecrane, line 75: What does the `prefix` do? Where does this value come from? Sorry still not super used to some of this syntax!

components/s3_bucket/main.tf
DFurnes: We have versioning enabled on our production bucket, so when we delete a file (or overwrite it, or whatnot) S3 keeps historical copies in case we need to roll back. This gives us extra peace of mind that we never lose user data from a bug! However, it'd also mean we'd keep these temporary files indefinitely unless we add a rule to remove those old versions too.

components/s3_bucket/main.tf
DFurnes: I made this change in d675657 (including removing other `id` values that should no longer be necessary).

components/s3_bucket/main.tf
katiecrane: Ah missed this comment on my first pass!

components/s3_bucket/main.tf
DFurnes: No worries, it's a little hard to follow!! The `prefix` argument "filters" which objects in the S3 bucket this rule applies to (so by default, it'd apply to everything & this argument limits it to only things that match a given file/folder prefix). The `lifecycle_rule.value` is the "iterator" for this dynamic block & comes from the `var.lifecycle_rule` input variable that we're iterating over.

DFurnes: So, in this case for Northstar, we're iterating over `["temporary/"]` and so we'd loop a single time where `lifecycle_rule.value` would be `"temporary/"` (and thus we'd only enforce these automatic deletion rules for files that were inside that folder).

katiecrane: So in the case of Northstar this limits to stuff in `temporary/`?

DFurnes: Yup! Here's what this turns into (from the plan) after Terraform does all it's fancy looping:

```hcl
lifecycle_rule {
  enabled = true
  prefix  = "temporary/"

  expiration {
    days = 14
  }

  noncurrent_version_expiration {
    days = 90
  }
}
```

DFurnes: Ideally we'd be able to just create this as a standalone `aws_s3_lifecycle_rule` resource in the Northstar module that looked exactly like that & deal with any looping funny-business to build this into the `s3_bucket` module but alas! 🙃

katiecrane: SO HELPFUL!! Thanks for spelling it out much much appreciated! I follow now!

katiecrane (approved):
Good from me! Thanks for teaching me!

components/s3_bucket/main.tf
DFurnes: Yay, that worked!! It automatically generated a `tf-s3-lifecycle-20210517211330135300000001` ID for this lifecycle rule when it created it. That makes these a lot easier to work with. 😌

